PR #2, opened 2022-01-25
Deploy

Kaitou786:


ui/public/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+  "short_name": "Canary Checker",
```
moshloop, line 2: We need to update these and other references from canary checker to Incident Commander

PR #17, opened 2022-03-29
chore: update helm chart values to reflect new canary-checker values

paddatrapper:


chart/values.yaml
```diff
@@ -14,17 +14,18 @@ image:
...
-    create: true
-    name: &secretKeyRef postgres-connection
-    configDbKey: &configDbKey CONFIG_DB_URL
-    canaryCheckerDbKey: &canaryCheckerKey CANARY_CHECKER_DB_URL
-    incidentCommanderDbKey: INCIDENT_COMMANDER_DB_URL
-  # Need to add the storage class and the storage according to the PV provisioned or according to the auto provitioning class that is configured on the cluster.
-  # In case no storageClass is defined the default storage class will be used.
-  storageClass:
-  storage: 10Gi
+  external:
```
moshloop, line 17: So we are anot moving incident-commander to external as well, we jist need to map the db parameters to db.external rather than db

paddatrapper: This keeps the DB external, as before. Just moves them under the correct key

moshloop: But it is in the wrong place - the incident commander db doesn't have the concept of an external db, only canary-checker, it needs to be move under `canary-checker.db.` not the root `db`

PR #58, opened 2022-06-01
chore: swicth to echo static middleware

Kaitou786:


ciju (approved):

cmd/server.go
```diff
@@ -30,10 +28,12 @@ var Serve = &cobra.Command{
 		forward(e, "/config", configDb)
 		forward(e, "/canary", canaryChecker)
 		forward(e, "/apm", apmHub)
-
-		contentHandler := echo.WrapHandler(http.FileServer(http.FS(ui.StaticContent)))
-		var contentRewrite = middleware.Rewrite(map[string]string{"/*": "/build/$1"})
-		e.GET("/*", contentHandler, contentRewrite)
+		e.Use(middleware.StaticWithConfig(middleware.StaticConfig{
+			Root:  "ui/build",
```
moshloop, line 32: I don't think we are copying the build directory into Docker image - and this would also break the binary - We would need to https://github.com/labstack/echo/blob/master/middleware/static.go#L48

Kaitou786: True.. updated

PR #61, opened 2022-06-05
feat: add middleware to add Cache-Control headers on response for sta…

Kaitou786:
…tic files


cmd/server.go
```diff
@@ -13,6 +14,19 @@ import (
...
+	CacheControlValue  = "max-age=600, public"
+)
+
+var cacheSuffixes = []string{
+	".ico",
+	".svg",
+	".css",
+	".js",
+	".png",
+}
```
Kaitou786, line 28: @moshloop let me know if we need to extend this list

cmd/server.go
```diff
@@ -13,6 +14,19 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	HeaderCacheControl = "Cache-Control"
+	CacheControlValue  = "max-age=600, public"
```
Kaitou786, line 19: Let me know if this value make sense to you

cmd/server.go
```diff
@@ -13,6 +14,19 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	HeaderCacheControl = "Cache-Control"
+	CacheControlValue  = "max-age=600"
```
Kaitou786, line 19: Let me know what value we need to define here as well... in the response

moshloop: public, max-age=2592000, immutable

moshloop (changes requested):

PR #82, opened 2022-07-12
feat: improve local setup by using docker compose

yashmehrotra:


Dockerfile
```diff
@@ -8,16 +8,14 @@ RUN  ls && npm run build
 
 FROM golang:1.17 as builder
 WORKDIR /app
-COPY ./ ./
```
yashmehrotra, line 11: Moved this below so that code changes does not cause a redownload of all dependencies

docker-compose.yml
```diff
@@ -0,0 +1,16 @@
...
+    environment:
+      DB_URL: postgres://ic:ic@db:5432/ic
+    command: serve
+  db:
+    image: postgres:14-alpine
+    restart: always
+    environment:
+      POSTGRES_DB: ic
+      POSTGRES_USER: ic
+      POSTGRES_PASSWORD: ic
```
yashmehrotra, line 16: Using `ic` (incident-commander) everywhere here for ease of typing, let me know if it is fine or should be changed

Dockerfile
```diff
@@ -8,28 +8,29 @@ RUN  ls && npm run build
...
-  apt-get install -y ca-certificates && \
+  apt-get install -y ca-certificates curl && \
   rm -Rf /var/lib/apt/lists/*  && \
   rm -Rf /usr/share/doc && rm -Rf /usr/share/man  && \
   apt-get clean
 
 COPY --from=builder /app/.bin/incident-commander /app
 
 RUN /app/incident-commander go-offline
+RUN curl -sSfL https://raw.githubusercontent.com/cosmtrek/air/master/install.sh | sh -s
```
moshloop, line 32: Do we need this if compose is using the air image? Shouldn't it be a in the make file as a dependency when running outside docker

yashmehrotra: Yeah, this is not needed. Removed it from dockerfile

Kaitou786:
@yashmehrotra do you think we can also add a compose for the entire stack?

With:

Canary-checker, config-db and apm-hum

yashmehrotra:
@Kaitou786 I'm not sure, ideally a compose for the entire stack should exist but not sure if it should be in this repo.

If I understand it correctly, flanksource-ui is the frontend which talks to multiple services (canary-checker, incident-commander, apm hub), but just for running/developing incident commander we do not need other services, just the UI.

But this is very subjective, I haven't worked on these long enough to have an opinion on the ideal flow here so would trust yours and @moshloop 's judgement this.

Kaitou786:
@yashmehrotra so incident-commander in itself doesn't provide much value, you can just create incidents but don't have anything to attach as evidence. 

The other micro-services can also provide value standalone, but from the incident-commander perspective I would say only the whole stack make sense and not incident-commander standalone
I also lean towards having the whole stack here, even if you look at the helm chart for the project that we have, we those micro-services as helm dependencies

moshloop:
I think it would make sense for the docker compose to spin up the other services as well, and maybe have an option run-dev-all that runs the binaries of the dependencies as well - Especially as we are merging the databases

moshloop:
I have also used something like this:

`Procfile`
```
canarywatch: sh -c 'cd /Users/moshe/go/src/github.com/flanksource/canary-checker &&   watchexec -e go -c make fast-build install'
configwatch: sh -c 'cd /Users/moshe/go/src/github.com/flanksource/confighub && watchexec -e go -c make build install'
apmwatch: sh -c 'cd /Users/moshe/go/src/github.com/flanksource/apm-hub && watchexec -e go -c make build install'
incidentwatch: sh -c 'cd /Users/moshe/go/src/github.com/flanksource/incident-commander && watchexec -e go -c make build install'
canary: canary-checker serve --httpPort 8081  --db    "postgres://localhost/incident_commander" -v
config: config-db serve --httpPort 8085  --db    "postgres://localhost/incident_commander" -v
ui: sh -c 'cd /Users/moshe/go/src/github.com/flanksource/flanksource-ui &&  npm run dev'
apm: apm-hub serve --httpPort 8082 /Users/moshe/go/src/github.com/flanksource/apm-hub/samples/config.yaml
incident: incident-commander serve --apm-hub http://localhost:8082 --canary-checker http://localhost:8081 --db    "postgres://localhost/incident_commander" --config-db http://localhost:8085
```

PR #83, opened 2022-07-13
feat: add basic functions for jira responder

yashmehrotra:


responder/jira/jira.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	if len(unsetVars) > 0 {
+		return fmt.Errorf("Required environment variables: %s", unsetVars)
+	}
+
+	return nil
+}
+
+func getClient() (*jira.Client, error) {
```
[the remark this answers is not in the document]
yashmehrotra: Cool 👍🏽

responder/jira/jira.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	client, err := jira.NewClient(tp.Client(), url)
+	if err != nil {
+		return nil, err
+	}
+	return client, nil
+}
+
+func CreateIssue(client *jira.Client, opts JiraOptions) error {
```
moshloop, line 34: this should be on the client itself e.g.  func (client *jira.Client) CreateIssue(issue JiraIssue) (string, error)

responder/jira/jira.go
```diff
@@ -0,0 +1,124 @@
...
+	"encoding/json"
+	"time"
+
+	"github.com/flanksource/commons/logger"
+	"github.com/flanksource/incident-commander/api"
+
+	jira "github.com/andygrunwald/go-jira"
+)
+
+type JiraOptions struct {
```
moshloop, line 13:
```suggestion
type JiraIssue struct {
```

I think we will need priority in here as well as maybe custom fields

responder/jira/jira.go
```diff
@@ -0,0 +1,124 @@
...
+	if err != nil {
+		return err
+	}
+	logger.Debugf("[Jira] Issue created with ID: [%s]", issue.ID)
+
+	return nil
+}
+
+func AddComment(client *jira.Client, issueID, comment string) error {
+
```
moshloop, line 62: gofmt the file, there shouldn't be newlines at the beginning or end of a method

responder/jira/jira.go
```diff
@@ -0,0 +1,124 @@
...
+	commObj := &jira.Comment{Body: comment}
+	c, _, err := client.Issue.AddComment(issueID, commObj)
+	if err != nil {
```
moshloop, line 65:
```suggestion
	c, _, err := client.Issue.AddComment(issueID, &jira.Comment{Body: comment})
	if err != nil {
```

responder/jira/jira.go
```diff
@@ -0,0 +1,124 @@
...
+			Body:      comment.Body,
+			CreatedBy: comment.Author.DisplayName,
+			CreatedAt: createdAt,
+		})
+	}
+
+	return comments, nil
+}
+
+func GetConfig(client *jira.Client) (json.RawMessage, error) {
```
moshloop, line 94: I think we want to create a struct and then marshall the struct

moshloop: i.e. GetProjects() []JiraProject

```
type JiraProject struct {
Key string
Name string
IssueTypes []string
}

File: responder/jira/jira.go (hunk 2 above)
Comment (yashmehrotra) [reply responder/jira/jira.go]: 👍🏽

File: responder/jira/jira.go (hunk 3 above)
Comment (yashmehrotra) [reply responder/jira/jira.go]: 👍🏽

File: responder/jira/jira.go [hunk 7]
```diff
@@ -0,0 +1,124 @@
...
+type JiraOptions struct {
+	ProjectKey  string
+	Summary     string
+	Description string
+	IssueType   string
+}
+
+func GetClient(email, apiToken, url string) (*jira.Client, error) {
+
+	tp := jira.BasicAuthTransport{
```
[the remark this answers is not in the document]
yashmehrotra: Got it

responder/jira/jira.go
yashmehrotra: I did gofmt it, it allows new lines, I'll remove them manually

responder/jira/jira.go
```diff
@@ -0,0 +1,124 @@
...
+			},
+			Summary: opts.Summary,
+		},
+	}
+
+	issue, _, err := client.Issue.Create(&i)
+	if err != nil {
+		return err
+	}
+	logger.Debugf("[Jira] Issue created with ID: [%s]", issue.ID)
```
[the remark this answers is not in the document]
yashmehrotra: 👍🏽

responder/jira/jira.go
yashmehrotra: 👍🏽

responder/jira/jira.go
yashmehrotra: 👍🏽

db/migrations/6_responder_queue.sql
```diff
@@ -0,0 +1,27 @@
...
+    error TEXT NULL,
+    created_at timestamp NOT NULL DEFAULT now(),
+    updated_at timestamp NOT NULL DEFAULT now()
+);
+
+CREATE OR REPLACE FUNCTION insert_responder_queue()
+RETURNS TRIGGER AS $$
+BEGIN
+    INSERT INTO responder_queue(responder_id) VALUES (new.id);
+    SELECT pg_notify("responder_updates", new.id);
```
moshloop, line 16: I don't think we can use pg_notify like this - It could cause updates to be duplicated, rather pg_notify should notify the go process to run a `select * from responder_queue with SKIP locked` e.g. https://www.enterprisedb.com/blog/what-skip-locked-postgresql-95 

That will prevent duplicate delivery and ensure all messages are delivered

yashmehrotra: Okay, interesting. We don't have to change anything here though, the query would be changed on the implementation end.

responder/jira/jira.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	return transitionList, err
+}
+
+func (jc JiraClient) TransitionIssue(issueID, transitionID string) error {
+	_, err := jc.client.Issue.DoTransition(issueID, transitionID)
+	if err != nil {
+		return err
+	}
+	return nil
```
moshloop, line 185:
```suggestion
	return err
```

responder/jira/jira.go
```diff
@@ -0,0 +1,186 @@
...
+
+	var transitionList []JiraIssueTransitions
+	for _, transition := range transitions {
+		transitionList = append(transitionList, JiraIssueTransitions{
+			ID:      transition.ID,
+			Name:    transition.Name,
+			ToState: transition.To.Name,
+		})
+	}
+	return transitionList, err
```
moshloop, line 177:
```suggestion
	return transitionList, nil
```

responder/jira/jira.go
```diff
@@ -0,0 +1,186 @@
...
+	for _, status := range statuses {
+		statusList = append(statusList, status.Name)
+	}
+
+	projects, _, err := jc.client.Project.GetList()
+	if err != nil {
+		return nil, err
+	}
+
+	p := make(map[string]JiraProject)
```
moshloop, line 139:
```suggestion
	projects := make(map[string]JiraProject)
```

responder/jira/jira.go
yashmehrotra: umm, this should be `nil` right ? cause if `err` we are returning that in the previous line

responder/jira/jira.go
```diff
@@ -0,0 +1,183 @@
...
+	for _, status := range statuses {
+		statusList = append(statusList, status.Name)
+	}
+
+	projects, _, err := jc.client.Project.GetList()
+	if err != nil {
+		return nil, err
+	}
+
+	projectsMap := make(map[string]JiraProject)
```
moshloop, line 139: nit: don't prefix or suffix variables with a type

PR #84, opened 2022-07-14
chore: update API's

moshloop:


api/types.go
```diff
@@ -86,16 +97,21 @@ type GenericTicketing struct {
 	Labels      map[string]string `json:"labels,omitempty"`
 }
 
+type JiraClient struct {
+	Url      string         `json:"url,omitempty"`
+	Username kommons.EnvVar `yaml:"username" json:"username"`
+	Password kommons.EnvVar `yaml:"password" json:"password"`
```
yashmehrotra, line 103: Should we use term `ApiToken` here since Jira discourages passwords and prefers using api tokens

moshloop: Doesn't the token just go into the password field ?

yashmehrotra: Yeah, the function/usage is same. I was just talking about the terminology (I had to scrap through the docs to make sure tokens can be used in the password field), but either of them is fine.

PR #89, opened 2022-07-19
include public endpoint config

ciju:


chart/templates/deployment.yaml
```diff
@@ -29,6 +29,7 @@ spec:
           args:
             - serve
             - -vvv
+            - --public-endpoint={{ .Values.ingress.host | quote }}
```
moshloop, line 32:
```suggestion
            - --public-endpoint=https://{{- .Values.ingress.host  }}
```

ciju: Does it make sense to let the browser and server handle the https case?. Server redirects http request. E.g 

```bash
$ curl -I incident-commander.canary.lab.flanksource.com
HTTP/1.1 308 Permanent Redirect
Date: Tue, 19 Jul 2022 15:27:16 GMT
Content-Type: text/html
Content-Length: 164
Connection: keep-alive
Location: https://incident-commander.canary.lab.flanksource.com
```

PR #93, opened 2022-07-25
feat: Add event processor for jira incidents

yashmehrotra:


events/events.go
```diff
@@ -0,0 +1,64 @@
...
+
+		reconcileEvents()
+
+	}
+}
+
+func reconcileEvents() error {
+	// TODO: Add attempts where clause
+	var events []api.Event
+	tx := db.Gorm.Raw("SELECT id, properties FROM event_queue FOR UPDATE SKIP LOCKED").Scan(&events)
```
moshloop, line 47: I think this needs to be done inside of a db transaction to work correctly, Without the transaction - the lock will be released in the next line.

events/events.go
```diff
@@ -0,0 +1,64 @@
...
+	if tx.Error != nil {
+		return tx.Error
+	}
+
+	var responderEvents []api.Event
+	for _, event := range events {
+		if event.Properties["type"] == "responder" {
```
moshloop, line 54: shouldn't this be queue name ?
I would process the events as is and not try to split them by type into a slicce and then process.

responder/events.go
```diff
@@ -0,0 +1,109 @@
...
+	// TODO: Temp solution to test workflow
+	// Get team details of all the responders
+	// team spec has jira details
+	/*
+		teamSpec := api.TeamSpec{
+			ResponderClients: api.ResponderClients{
+				Jira: api.JiraClient{
+					Url:      "https://flanksource.atlassian.net",
+					Username: kommons.EnvVar{Value: "yash@flanksource.com"},
+					Password: kommons.EnvVar{Value: "7NCOe92hvnKZBzjenSOnD673"},
```
moshloop, line 83: You might need to reset you password....

responder/events.go
```diff
@@ -0,0 +1,109 @@
...
+				},
+			},
+		}*/
+
+	jc, err := jira.NewClient(
```
moshloop, line 88:
```suggestion
	jiraClient, err := jira.NewClient(
```
or just client

responder/events.go
```diff
@@ -0,0 +1,109 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	issue, err := jc.CreateIssue(issueOptions)
+
+	if err != nil {
+		return "", err
+	}
+	return issue.Key, nil
```
moshloop, line 108:
```suggestion
	return issue.Key, err
```

responder/events.go
```diff
@@ -0,0 +1,109 @@
...
+			}
+
+			// Cautionary assignment to prevent previous external_id being set in database
+			externalID = ""
+		}
+
+		err := event.Done()
+		if err != nil {
+			logger.Errorf("Error deleting event from event_queue table:responder with id:%s and external_id:%s", responder.ID, externalID)
+		}
```
moshloop, line 62:
```suggestion
		if err := event.Done(); err != nill {
			logger.Errorf("Error deleting event from event_queue table:responder with id:%s and external_id:%s", responder.ID, externalID)
		}
```

responder/events.go
```diff
@@ -0,0 +1,109 @@
...
+	var responderIDs []string
+	responderIDEventMap := make(map[string]api.Event)
+	for _, event := range events {
+		responderIDs = append(responderIDs, event.Properties["id"])
+		responderIDEventMap[event.Properties["id"]] = event
+	}
+
+	var responders []api.Responder
+
+	tx := db.Gorm.Find(&responders).Where("id IN (?) AND external_id is NULL", responderIDs).Preload("Team")
```
moshloop, line 26: I don't think this can return more than 1 entry

responder/events.go
moshloop: I think move the error handling into event.Done() - longer term we might want to add some retry logic etc..

responder/events.go
```diff
@@ -0,0 +1,109 @@
...
+				Jira: api.JiraClient{
+					Url:      "https://flanksource.atlassian.net",
+					Username: kommons.EnvVar{Value: "yash@flanksource.com"},
+					Password: kommons.EnvVar{Value: "7NCOe92hvnKZBzjenSOnD673"},
+				},
+			},
+		}*/
+
+	jc, err := jira.NewClient(
+		teamSpec.ResponderClients.Jira.Username.Value,
```
moshloop, line 89: We need to use kommons.GetEnvValueFromCache here

responder/events.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/flanksource/commons/logger"
+	_ "github.com/flanksource/kommons"
+	"github.com/mitchellh/mapstructure"
+
+	"github.com/flanksource/incident-commander/api"
+	"github.com/flanksource/incident-commander/db"
+	"github.com/flanksource/incident-commander/responder/jira"
+)
+
+func ReconcileEvents(events []api.Event) {
```
moshloop, line 15: I think move most of this into events package and keep all the event handling code together.

responder/events.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+
+		err := event.Done()
+		if err != nil {
+			logger.Errorf("Error deleting event from event_queue table:responder with id:%s and external_id:%s", responder.ID, externalID)
+		}
+	}
+}
+
+func handleJiraResponder(responder api.Responder) (string, error) {
```
moshloop, line 66:
```suggestion
func notifyJiraResponder(responder api.Responder) (string, error) {
```

responder/events.go
```diff
@@ -0,0 +1,109 @@
...
+			externalID, err = handleJiraResponder(responder)
+			if err != nil {
+				setErr := event.SetErrorMessage(err.Error())
+				if setErr != nil {
+					logger.Errorf("Error updating table:event_queue with id:%s and error:%s", event.ID, err)
+				}
+				continue
+			}
+		}
+
```
moshloop, line 47: refactor this into into a notifyResponder function annd leave behind in the responder package

events/events.go
moshloop: It also needs to `limit 1` otherwise if there is an error processing a single item, then the entire transaction for the batch is rolled bacl. 

I would maybe put this into a loop, that selects, if nothing found then sleep for configurable time period, and then  use a channel to wake from the sleep 

```go
select {
	case m := <-pgNotify:
		handle(m)
	case <-time.After(10 * time.Second):
		fmt.Println("timed out")
	}
```

events/events.go
```diff
@@ -0,0 +1,64 @@
...
+	//"fmt"
+
+	"github.com/flanksource/commons/logger"
+
+	"github.com/flanksource/incident-commander/api"
+	"github.com/flanksource/incident-commander/db"
+	"github.com/flanksource/incident-commander/responder"
+)
+
+func ProcessQueue() error {
```
moshloop, line 14:
```suggestion
func ListenForEvents() error {
```

events/events.go
```diff
@@ -0,0 +1,64 @@
...
+		if err != nil {
+			return err
+		}
+
+		reconcileEvents()
+
+	}
+}
+
+func reconcileEvents() error {
```
moshloop, line 44:
```suggestion
func consumeEvents() error {
```

responder/events.go
yashmehrotra: Ahh, shucks, I removed the token but forgot to commit that, have revoked it now

responder/events.go
yashmehrotra: @moshloop If there is an error in the issue creation, the `issue.Key` won't be set and will give a `invalid memory address or nil pointer dereference` error

moshloop: ok, then follow the convention of `if err := doSomthing(); err != nil`

events/events.go
```diff
@@ -0,0 +1,137 @@
...
+			logger.Errorf("Error waiting for database notifications: %v", err)
+		}
+
+		pgNotify <- true
+	}
+
+}
+
+// TODO: Better function name
+func consumeEventsWrapper() {
```
yashmehrotra, line 62: @moshloop @Kaitou786 @ciju 

Need better name suggestions, context is that this function will call `consumeEvents` (which will be renamed to `consumeEvent`) infinitely till there are no more events to consume

api/types.go
```diff
@@ -19,6 +24,15 @@ type Hypothesis struct {
 }
 
 type Responder struct {
+	ID         uuid.UUID
+	Properties types.JSONStringMap `json:"properties" gorm:"type:jsonstringmap;<-:false"`
+	TeamID     uuid.UUID
+	// TODO: Impement Scan+Value for kommons.EnvVar for gorm to work correctly
```
yashmehrotra, line 30: @moshloop Has this been implemented somewhere or should I create an issue in the kommons repo and add this there ?

Basically, for gorm to use kommons.EnvVar for storage and query we need to implement methods similar to jsonStringMap

events/events.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+}
+
+func consumeEvents() error {
+	var event api.Event
+
+	tx := db.Gorm.Begin()
+
+	// TODO: Add attempts where clause
+	err := tx.Raw("SELECT id, name, properties FROM event_queue FOR UPDATE SKIP LOCKED LIMIT 1").First(&event).Error
```
yashmehrotra, line 82: In case of failed events, we are adding an error message and incrementing the attempt count.

But I feel we should not retry/pick an event which has >= N attempts. What should N be ? I was thinking 3 attempts are more than enough ?

events/events.go
```diff
@@ -0,0 +1,137 @@
...
+		return tx.Commit().Error
+	}
+
+	err = tx.Delete(&event).Error
+	if err != nil {
+		logger.Errorf("Error deleting event from event_queue table with id:%s", event.ID.String())
+		// TODO: In this case, the event has been processed successfully, but its deletion is failing
+		// Rolling back would lead to a dangling responder and may result in duplication
+		// But commiting the transaction would lead to the event being processed again, ie. duplication
+		// Choose lesser of the two evils
```
yashmehrotra, line 106: Thoughts on how to handle this ?

responder/events.go
yashmehrotra: We would need to create a kubernetes client for using `kommons.GetEnvValueFromCache`

I thought we were going with the approach of -> load crds into database -> query/treat them as normal jsons/structs

api/types.go
moshloop: So we wouldn;t be persisting envVar to the db, we are marshalling the entire struct and then saving it as json

events/events.go
moshloop: I would just merge this function into consumeEvents

events/events.go
moshloop: I think we should always retry, and then add an explicit action to flush failed attempts, the problem with stopping to to retry is the connection could have an expired token/password that takes a couple days to fix and by then you would need to perform some other manual action other than updating the password. 

We should be backing off, i.e. retry after 1, 5, 15, 60m, and there after every 3h

events/events.go
moshloop: In all likelihood if the delete fails, the rollback will fail as well and/or the transaction will timeout - This is a very rare edge case that I don't think we need to handle for now - most likely long term we would implement a real queuing system that handles these types of issues.

responder/events.go
moshloop: Yes,  but in most cases we are still running in kubernetes and we don't want to store the secrets in the db, so we can look them up at runtime via kubernetes secrets.

events/events.go
yashmehrotra: It was previously in the same function but it was a bit complex to read and understand, so decided to split it. I'll use a nested function

db/migrations/005_events.sql
```diff
@@ -4,7 +4,7 @@
 -- Event queue for events that need to be processed in the background, once succesfully processed the event is removed from the queue
 CREATE TABLE event_queue (
     id UUID DEFAULT generate_ulid() PRIMARY KEY,
-    event TEXT NOT NULL,
+    name TEXT NOT NULL,
```
moshloop, line 7: Wouldn't this need to be a migration now ? unless we are dropping the bd again ?

yashmehrotra: I assumed it will be merged together with the db migration fix, so would have to only drop/recreate once

PR #104, opened 2022-08-08
feat: MS planner responder integration

yashmehrotra:
Opened a draft PR for early suggestions/feedback. Few parts are yet to be completed.


responder/msplanner/msplanner.go
```diff
@@ -0,0 +1,196 @@
...
+		convBody := models.NewConversationThread()
+		topic := fmt.Sprintf("Conversation thread topic for taskID: %s", taskID)
+		convBody.SetTopic(&topic)
+
+		convBody.SetPosts([]models.Postable{post})
+
+		result, err := c.client.GroupsById(c.groupID).Threads().Post(convBody)
+		if err != nil {
+			fmt.Println("error creating Conversation thread")
+			return openDataError(err)
```
moshloop, line 106: Don't log and return an error

responder/msplanner/msplanner.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+
+		// TODO: For debugging
+		fmt.Println("Conversation thread created")
+		fmt.Println(result)
```
moshloop, line 111: use commons/logger trace

responder/msplanner/msplanner.go
yashmehrotra: Yeah, this is just for debugging on local, will remove all print statements

responder/msplanner/msplanner.go
yashmehrotra: Yeah, this is just for debugging on local, will remove all print statements

events/events.go
```diff
@@ -127,9 +127,12 @@ func reconcileResponderEvent(tx *gorm.DB, event api.Event) error {
 	var externalID string
 	if responder.Properties["responderType"] == "Jira" {
 		externalID, err = responderPkg.NotifyJiraResponder(responder)
-		if err != nil {
-			return err
-		}
+	} else if responder.Properties["responderType"] == "MSPlanner" {
+		externalID, err = responderPkg.NotifyMSPlannerResponder(responder)
+	}
+
```
yashmehrotra, line 133: @moshloop Should I use a `getResponderFunc(responderType string) func(responder) (string, error)` or a `switch` here or is if/else fine for now

events/events.go
```diff
@@ -127,9 +127,12 @@ func reconcileResponderEvent(tx *gorm.DB, event api.Event) error {
 	var externalID string
 	if responder.Properties["responderType"] == "Jira" {
 		externalID, err = responderPkg.NotifyJiraResponder(responder)
-		if err != nil {
-			return err
-		}
+	} else if responder.Properties["responderType"] == "MSPlanner" {
```
Kaitou786, line 130: Can we make these `MSPlanner` and JIRA as constants and probably create a type Responders which is a string and restrict the types from there?

responder/events.go
```diff
@@ -48,3 +50,43 @@ func NotifyJiraResponder(responder api.Responder) (string, error) {
...
+	if err = json.Unmarshal(teamSpecJson, &teamSpec); err != nil {
+		return "", err
+	}
+
+	client, err := msplanner.NewClient(
+		teamSpec.ResponderClients.MSPlanner.TenantID,
+		teamSpec.ResponderClients.MSPlanner.ClientID,
+		teamSpec.ResponderClients.MSPlanner.GroupID,
+		teamSpec.ResponderClients.MSPlanner.Username.Value,
+		teamSpec.ResponderClients.MSPlanner.Password.Value,
```
Kaitou786, line 74: Will we be getting the value directly or do we need to first fetch it from a secret?
Like we have in canary-checker at `checks/common.go:L65`

responder/msplanner/msplanner.go
```diff
@@ -0,0 +1,214 @@
...
+			Name:       *plan.GetTitle(),
+			Buckets:    buckets,
+			Priorities: priorities,
+		}
+	}
+
+	return config, nil
+}
+
+func openDataError(err error) error {
```
Kaitou786, line 196: just curious why do we need to cater for it specially

events/events.go
Kaitou786: I would incline towards using a switch

responder/msplanner/msplanner.go
yashmehrotra: The microsoft graph API returns specific `openData` errors where you have two extract multiple things from the error. If I just log the error as is, we'll get a `error bad status code received`

events/events.go
yashmehrotra: 👍🏽

responder/events.go
yashmehrotra: We'll be fetching the secret and storing it in the database, and the database would be the source of truth. So this flow assumes that the k8s object -> database part is already done.

PR #106, opened 2022-08-10
chore: fix lint timeouts

yashmehrotra:


yashmehrotra:
So using the `staticcheck` linter fails the entire lint run

PR #111, opened 2022-08-16
Add comments from incidents to responders

yashmehrotra:


events/events.go
```diff
@@ -146,3 +157,57 @@ func reconcileResponderEvent(tx *gorm.DB, event api.Event) error {
...
+			// For now, we are only logging the error and moving on
+			// In future, we would want to associate error messages with responderType
+			// and when reprocessing, handle it only for specific responders
+			logger.Errorf("Error adding comment to responder:%s %v", responder.Properties["responderType"], err)
+			continue
+		}
+
+		// Insert into comment_responders table
+		if externalID != "" {
+			insertQuery := `
```
moshloop, line 201: Why not just inline the query ?

yashmehrotra: primarily for readability, keeping the query and the statement separate

events/events.go
```diff
@@ -146,3 +157,57 @@ func reconcileResponderEvent(tx *gorm.DB, event api.Event) error {
...
+	var responders []api.Responder
+	commentRespondersQuery := `
+        SELECT * FROM responders WHERE incident_id IN (
+            SELECT incident_id FROM comments WHERE id = ?
+        )
+    `
+	err := tx.Raw(commentRespondersQuery, commentID).Preload("Team").Find(&responders).Error
+	if err != nil {
+		return err
+	}
```
moshloop, line 175:
```suggestion
	if err := tx.Raw(commentRespondersQuery, commentID).Preload("Team").Find(&responders).Error; err != nil {
		return err
	}
```

events/events.go
```diff
@@ -146,3 +157,57 @@ func reconcileResponderEvent(tx *gorm.DB, event api.Event) error {
...
+			externalID, err = responderPkg.NotifyJiraResponderAddComment(responder, commentBody)
+		case responderPkg.MSPlannerResponder:
+			externalID, err = responderPkg.NotifyMSPlannerResponderAddComment(responder, commentBody)
+		default:
+			continue
+		}
+
+		if err != nil {
+			// For now, we are only logging the error and moving on
+			// In future, we would want to associate error messages with responderType
```
moshloop, line 193:
```suggestion
			// TODO associate error messages with responderType
```

events/events.go
moshloop: I think it makes it less readable, if it is a very long query that is one thing, but if the query can fit on single line, and inlined in the parameters its fewer lines to read

responder/events.go
```diff
@@ -96,3 +118,54 @@ func NotifyMSPlannerResponder(responder api.Responder) (string, error) {
...
+	if err != nil {
+		return "", err
+	}
+
+	err = client.AddComment(responder.ExternalID, comment)
+
+	// MS Graph API does not return the ID of the comment created
+	// so we just return a constant string for now
+	return "posted", err
+}
```
moshloop, line 171: I think all of this should go into the msplanner package,  and then maybe add a registry of responderType to responder interface

api/types.go
```diff
@@ -10,12 +10,18 @@ import (
...
 type Comment struct {
-	Body      string    `json:"body"`
-	CreatedBy string    `json:"created_by"`
-	CreatedAt time.Time `json:"created_at"`
+	Body       string    `json:"body"`
+	CreatedBy  string    `json:"created_by"`
+	CreatedAt  time.Time `json:"created_at"`
+	IncidentID uuid.UUID `json:"incident_id"`
+	Incident   Incident  `json:"incident"`
 }
```
moshloop, line 25: Wdo we have both?

events/events.go
yashmehrotra: Cool

events/events.go
yashmehrotra: 👍🏽

events/events.go
yashmehrotra: 👍🏽

responder/events.go
yashmehrotra: Yeah, moving it to a package is better, I'll pick up the interface part in the config task

api/types.go
yashmehrotra: yeah, removing `incident_id` for now

db/migrations/004_incidents.sql
```diff
@@ -89,16 +89,24 @@ CREATE TABLE comments (
   id UUID DEFAULT generate_ulid() PRIMARY KEY,
   created_by UUID NOT NULL ,
   comment text NOT NULL,
-  external_id TEXT NULL, -- A unique identifier for the responder in the external system e.g. Jira ticket id
   incident_id UUID NOT NULL,
-  responder_id UUID NULL,
```
moshloop, line 94: I think we still need a responder_id, this will be the source of the comment

yashmehrotra: Oh, for comments made from responder and added here ? I guess so, in that case we should put back the external_id as well here.

api/types.go
```diff
@@ -10,12 +10,17 @@ import (
 )
 
 type Incident struct {
+	ID     uuid.UUID `json:"id"`
+	Title  string    `json:"title"`
+	TeamID uuid.UUID `json:"team_id"`
+	Team   Team      `json:"team"`
```
moshloop, line 16: Why both ?

yashmehrotra: I initially thought there maybe usecases where we may need teamID but not the entire team object but I don't think thats the case for now, have removed it from here and other places as well

PR #120, opened 2022-08-22
feat: job for syncing config

yashmehrotra:


responder/config_sync.go
```diff
@@ -0,0 +1,83 @@
...
+				continue
+			}
+
+			jiraConfigJSON, err := jiraClient.GetConfigJSON()
+			if err != nil {
+				logger.Errorf("Error generating config from Jira: %v", err)
+				continue
+			}
+
+			if err = db.Gorm.Exec(dbInsertConfigQuery, JiraResponder, "Jira", team.ID, jiraConfigJSON).Error; err != nil {
```
yashmehrotra, line 56: @moshloop What should we put in the `name` column ? I'm using the responder name and team id for config_type  and external_id respectively, unclear on what the `name` should be

responder/config_sync.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/flanksource/incident-commander/api"
+	"github.com/flanksource/incident-commander/db"
+)
+
+func StartConfigSync() {
+	for {
+		logger.Infof("Syncing configuration")
+		syncConfig()
+
+		time.Sleep(10 * time.Minute)
```
yashmehrotra, line 18: Is 10 minutes good enough ?

moshloop: Its probably too often - I think change to 1h, and then when we build the integration table we can add a schedule

responder/config_sync.go
moshloop: The name would be the URL I think of the jira instance

responder/config_sync.go
```diff
@@ -0,0 +1,102 @@
...
+		time.Sleep(1 * time.Hour)
+	}
+}
+
+func upsertConfig(configType, externalID, name, config string) error {
+
+	dbInsertConfigQuery := `INSERT INTO config_items (config_type, name, external_id, config) VALUES (?, ?, ARRAY[?], ?)`
+	dbUpdateConfigQuery := `UPDATE config_items SET config = ? WHERE external_id = ARRAY[?] AND config_type = ?`
+
+	query := db.Gorm.Exec(dbUpdateConfigQuery, config, externalID, configType)
```
moshloop, line 27: nit: s/query/tx/

responder/config_sync.go
```diff
@@ -0,0 +1,102 @@
...
+
+func syncConfig() {
+	var teams []api.Team
+	if err := db.Gorm.Find(&teams).Error; err != nil {
+		logger.Errorf("Error querying teams from database: %v", err)
+		return
+	}
+
+	for _, team := range teams {
+		teamSpecJson, err := team.Spec.MarshalJSON()
```
moshloop, line 51: I would move this into a function e.g. api.TeamFromV1(team.spec) or team.GetSpec()

PR #133, opened 2022-08-26
Sync comments from jira to incident commander

yashmehrotra:


api/types.go
```diff
@@ -29,6 +34,9 @@ type Responder struct {
 	ID         uuid.UUID           `json:"id"`
 	Properties types.JSONStringMap `json:"properties" gorm:"type:jsonstringmap;<-:false"`
 	ExternalID string              `json:"external_id"`
+	IncidentID uuid.UUID           `json:"incident_id"`
```
moshloop, line 37: We need a better solution for having both incidentid and incident - but its ok for now

PR #136, opened 2022-08-30
feat: add component selection logic for teams

Kaitou786:


db/components.go
```diff
@@ -0,0 +1,115 @@
...
+
+	"github.com/flanksource/commons/logger"
+	"github.com/flanksource/incident-commander/api"
+	"github.com/flanksource/incident-commander/db/types"
+	"github.com/flanksource/incident-commander/utils"
+	"github.com/google/uuid"
+	"gorm.io/gorm/clause"
+)
+
+func GetTeamsWithComponentSelector() map[uuid.UUID][]api.ComponentSelector {
```
yashmehrotra, line 15: Lets move this to a `teams` package?

db/components.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+	}
+}
+
+func PersistTeamComponent(teamComp api.TeamComponent) error {
+	tx := Gorm.Clauses(clause.OnConflict{
+		Columns:   []clause.Column{{Name: "team_id"}, {Name: "component_id"}, {Name: "selector_id"}},
+		UpdateAll: true,
+	}).Create(teamComp)
+	return tx.Error
```
yashmehrotra, line 97: (nitpick): `return db.Gorm...Error`

utils/utils.go
```diff
@@ -0,0 +1,19 @@
+package utils
+
```
yashmehrotra, line 2: lets rename it to `utils/hash.go`

db/components.go
```diff
@@ -0,0 +1,115 @@
...
+		args["labels"] = types.JSONStringMap(selector.Labels)
+	}
+	rows, err := QueryNamed(context.Background(), sql, args)
+	if err != nil {
+		logger.Errorf("error fetching component with selector: %v", selector)
+	}
+	for rows.Next() {
+		var id uuid.UUID
+		if err := rows.Scan(&id); err != nil {
+			logger.Errorf("error scanning the id for the component")
```
moshloop, line 76: rather return this

utils/utils.go
yashmehrotra: Reasoning: https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common

db/components.go
```diff
@@ -0,0 +1,115 @@
...
+	if selector.Type != "" {
+		sql += " AND type = :type"
+		args["type"] = selector.Type
+	}
+
+	if selector.Labels != nil {
+		sql += "AND labels @> :labels"
+		args["labels"] = types.JSONStringMap(selector.Labels)
+	}
+	rows, err := QueryNamed(context.Background(), sql, args)
```
moshloop, line 69: We might be able to replace this with: https://gorm.io/docs/advanced_query.html#Named-Argument

db/migrations/003_teams.sql
```diff
@@ -45,11 +45,15 @@ CREATE TABLE  IF NOT EXISTS team_components (
...
   role TEXT NULL,
+  selector_id TEXT,
   PRIMARY KEY (team_id, component_id),
   FOREIGN KEY (team_id) REFERENCES teams(id)
-  -- FOREIGN KEY (component_id) REFERENCES components(id)
+  FOREIGN KEY (component_id) REFERENCES components(id)
+  UNIQUE (team_id, component_id, selector_id)
 );
 
+INSERT into incident_commander_db_version(version_id, tstamp, is_applied) VALUES(6, NOW(), true);
```
yashmehrotra, line 55: We can hold off the metadata increment handling till we move to v1 or ship to a customer, what to you think ? Saying it because the schema's are in constantly changing state till the MVD is ready

db/components.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/google/uuid"
+	"gorm.io/gorm/clause"
+)
+
+func GetTeamsWithComponentSelector() map[uuid.UUID][]api.ComponentSelector {
+	var teams []api.Team
+	var teamComponentMap = make(map[uuid.UUID][]api.ComponentSelector)
+	err := Gorm.Table("teams").Where("spec::jsonb ? 'components';").Find(&teams).Error
+	if err != nil {
+		logger.Errorf("error fetching the teams with componenets: %v", err)
```
yashmehrotra, line 20: return teamComponentMap

db/components.go
```diff
@@ -0,0 +1,115 @@
...
+	for _, team := range teams {
+		teamSpec := &api.TeamSpec{}
+		teamSpecJson, err := team.Spec.MarshalJSON()
+		if err != nil {
+			logger.Errorf("error marshalling team spec for team: %v", team.ID)
+			continue
+		}
+		if err := json.Unmarshal(teamSpecJson, teamSpec); err != nil {
+			logger.Errorf("error unmarshalling the teamSpec for team: %v", team.ID)
+		}
```
yashmehrotra, line 31: you could use `team.GetSpec()` for this

db/components.go
```diff
@@ -0,0 +1,115 @@
...
+	var selectedComponents = make(map[string][]uuid.UUID)
+	for _, compSelector := range componentSelectors {
+		selectedComponents[utils.GetHash(compSelector)] = getComponentsWithSelector(compSelector)
+	}
+
+	return selectedComponents
+}
+
+func getComponentsWithSelector(selector api.ComponentSelector) []uuid.UUID {
+	sql := "select ID from components where deleted_at is null "
```
yashmehrotra, line 47: use `query` variable since `database/sql` is a stdlib package and we are using it in this project

db/components.go
yashmehrotra, line 76: add `%v, err` for more verbosity

db/init.go
```diff
@@ -104,12 +107,35 @@ func Migrate() error {
...
+
+func ConvertNamedParams(sql string, namedArgs map[string]interface{}) (string, []interface{}) {
+	i := 1
+	var args []interface{}
+	// Loop the named args and replace with placeholders
+	for pname, pval := range namedArgs {
+		sql = strings.ReplaceAll(sql, ":"+pname, fmt.Sprint(`$`, i))
+		args = append(args, pval)
+		i++
+	}
```
yashmehrotra, line 134: Is it possible to use `sql.Named` here ? It works something like this 
```golang
db.ExecContext(ctx, `
    delete from Invoice
    where
        TimeCreated < @end
        and TimeCreated >= @start;`,
    sql.Named("start", startTime),
    sql.Named("end", endTime),
)
```

db/init.go
```diff
@@ -104,12 +107,35 @@ func Migrate() error {
...
+		}
 	}
 	return nil
 }
 
 func GetDB() (*sql.DB, error) {
 	return sql.Open("pgx", pgxConnectionString)
 }
+
+func ConvertNamedParams(sql string, namedArgs map[string]interface{}) (string, []interface{}) {
```
yashmehrotra, line 126: Also, lets use the variable `query` here

db/components.go
```diff
@@ -0,0 +1,115 @@
...
+		sql += " AND namespace = :namespace"
+		args["namespace"] = selector.Namespace
+	}
+
+	if selector.Type != "" {
+		sql += " AND type = :type"
+		args["type"] = selector.Type
+	}
+
+	if selector.Labels != nil {
```
yashmehrotra, line 65: Could we use chaining here ? https://gorm.io/docs/method_chaining.html

```golang

query = db.Model(Components).Where()

if namespace ! = {
  query = query.Where("namespace  = ?", namespace)
}

if selector != "" {
  query = query.Where("type = "", type)
}
```

and so on, seems cleaner and easier to read

db/migrations/003_teams.sql
moshloop: I think thats fine - I want to avoid having to recreate the db - Just before v1 we can refactor down

db/components.go
Kaitou786: Yeah this make more sense

db/init.go
Kaitou786: We don't need this functions since went ahead with the chaining

db/init.go
Kaitou786: Same function is removed in favour of chained query

db/components.go
Kaitou786: Not applicable any more.. since code removed

db/components.go
Kaitou786: not applicable anymore

db/components.go
Kaitou786: Yeah make sense... added a function for it

db/components.go
Kaitou786: missed that. thanks!

db/components.go
Kaitou786: Using Chained Queries for now. As Yash mentioned easier to do and read

utils/utils.go
Kaitou786: done

db/components.go
Kaitou786: NA

db/components.go
Kaitou786: Not sure how that would be different than tx.Error?

db/components.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+	return teamComponentMap
+}
+
+func GetComponentsWithSelectors(componentSelectors []api.ComponentSelector) map[string][]uuid.UUID {
+	var selectedComponents = make(map[string][]uuid.UUID)
+	for _, compSelector := range componentSelectors {
+		selectedComponents[utils.GetHash(compSelector)] = getComponentsWithSelector(compSelector)
+	}
+	fmt.Println(selectedComponents)
```
yashmehrotra, line 38: Remove

db/components.go
Kaitou786: Created a teams package for the non-db related things. thanks!

PR #152, opened 2022-09-14
add chart dependencies for front-end

ciju:
Helm diff for prod: https://gist.github.com/ciju/cf4dbef6ea71716ec3b7ea2543c2aed7

Shows more than actual diff, because of the diff tool.


chart/values.yaml
```diff
@@ -21,57 +17,127 @@ db:
...
     kubernetes.io/ingress.class: nginx
     # kubernetes.io/tls-acme: "true"
-  host: chart-example.local
+  host: incident-commander.canary.lab.flanksource.com
   tls: []
   #  - secretName: chart-example-tls
   #    hosts:
   #      - chart-example.local
 
+canary-checker-ui:
```
moshloop, line 49: We don't need this

chart/values.yaml
```diff
@@ -21,57 +17,127 @@ db:
...
+      key: DB_URL
+
+# Enable ingress only if the UI is deployed outside of the cluster and calls public incident-commander api endpoint.
 ingress:
   enabled: false
   annotations:
     kubernetes.io/ingress.class: nginx
     # kubernetes.io/tls-acme: "true"
-  host: chart-example.local
+  host: incident-commander.canary.lab.flanksource.com
```
moshloop, line 43: It shouldn't default to this

chart/templates/secrets.yaml
```diff
@@ -0,0 +1,42 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "kratos-im.secretname" . }}
+  namespace: {{ .Release.Namespace }}
+type: Opaque
+data:
+  dsn: {{ (print "postgres://" $user ":" $password "@" $host "/incident_commander") | b64enc | quote }}
+  # Generate a random secret if the user doesn't give one. User given secret has priority
+  secretsDefault: {{ ( include "kratos-im.secrets.default" . | default ( randAlphaNum 32 )) | required "Value kratos.config.secrets.default can not be empty!" | b64enc | quote }}
```
moshloop, line 35: Does this not get regenerated every time the chart is applied ?

chart/Chart.yaml
```diff
@@ -14,3 +14,19 @@ dependencies:
   - name: canary-checker
     version: "0.38.158"
     repository: https://flanksource.github.io/charts
+  - name: flanksource-ui
+    version: "1.0.117"
+    repository: https://flanksource.github.io/charts
+    condition: canary-checker-ui.enabled
+    alias: canary-checker-ui
```
moshloop, line 21:
```suggestion
```

chart/files/kratos.yaml
```diff
@@ -0,0 +1,72 @@
...
+  secrets:
+    default:
+      - lorem ipsum dolores
```
moshloop, line 71: I don't think we should be adding these defaults

chart/values.yaml
```diff
@@ -1,14 +1,10 @@
...
-# Declare variables to be passed into your templates.
-
 replicas: 1
 
 # Use this only if you want to replace the default that is .Chart.Name as the name of all the objects.
 nameOverride: ""
 
 image:
-  repository: docker.io/flanksource/incident-commander
+  repository: localhost:5001/incident-commander
```
moshloop, line 7:
```suggestion
  repository: docker.io/flanksource/incident-commander
```

chart/files/kratos.yaml
ciju: This  specific file is unused. Will delete it. Similar config is there in files/kratos-config.yaml. But it can be overwritten from values.yaml. Should I remove the secrets from files/kratos-config.yaml?

chart/Chart.yaml
ciju: This is for existing deployments of canary-checker we have. Otherwise we won't be able to deploy updates to existing canary-checker deployments. \cc @Kaitou786

chart/templates/secrets.yaml
ciju: ack. had hooks earlier. Will update it to generate only if no existing password (similar to postgress secrets),

Issue #154, opened 2022-09-14
Update config sync for better visibility 


yashmehrotra:
- Change jira, ms_planner to `Jira` and `MSPlanner` in type
- Along with team_id, add board_id or plan_id in external_id as well

Issue #153, opened 2022-09-14
Add support for defaults and values in team spec's responders


yashmehrotra:
In the team's spec, 
```json
{
  "responders": {
     "jira": {
        "values": {"project": ""},  // Cannot be modified
        "defaults": {"issueType": ""}, // Can be modified but set as the default
      }
}
```

PR #156, opened 2022-09-15
feat: Add support for defaults and values in team spec

### LINKED ISSUE #150: MS Planner config sync - Use Plan/Board name
It should only create a single config_item per board and then update it on subsequent runs. 
Plan id should be the external_id for deduplication

yashmehrotra:
Fixes #150, #153, #154

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -83,10 +83,6 @@ type Slack struct {
 	Channel      string `json:"channel"`
 }
 
-type ResponderValues map[string]string
-
-type ResponderDefaults map[string]string
-
 type ResponderClients struct {
 	Jira      *JiraClient      `json:"jira,omitempty"`
 	AWS       *AWSClient       `json:"aws,omitempty"`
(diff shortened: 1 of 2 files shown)


api/types.go
```diff
@@ -154,17 +158,21 @@ type GenericTicketing struct {
 }
 
 type JiraClient struct {
-	Url      string         `json:"url,omitempty"`
-	Username kommons.EnvVar `yaml:"username" json:"username"`
-	Password kommons.EnvVar `yaml:"password" json:"password"`
+	Url      string            `json:"url,omitempty"`
+	Username kommons.EnvVar    `yaml:"username" json:"username"`
+	Password kommons.EnvVar    `yaml:"password" json:"password"`
+	Values   ResponderValues   `json:"values"`
```
moshloop, line 164: I think we can put this into a common struct, and then embed with `json:",inline"` - also no need to create a custom type for the map[string]string

PR #158, opened 2022-09-22
feat: add evidence types and refactor db

moshloop:


db/migrations/004_incidents.sql
```diff
@@ -137,13 +150,27 @@ CREATE TABLE evidences (
...
   hypothesis_id UUID NOT NULL,
+  config_id UUID NULL,
+  config_change_id UUID null,
+  config_analysis_id UUID null,
+  component_id UUID null,
+  check_id UUID null,
+  definition_of_done boolean, -- This indicates this item as needing to be fixed before closing the incident
+  done boolean, -- The evidence is done / resolved
+  factor boolean,
+  mitigator boolean,
```
yashmehrotra, line 161: Should we store these in a json format ? Is there a possibility that more similar fields would be added ?

db/init.go
```diff
@@ -82,17 +101,42 @@ func Init(connection string) error {
...
-	if err != nil {
+	if err = Migrate(); err != nil {
 		return err
 	}
 
-	return Migrate()
+	system := api.Person{}
+	if err = Gorm.Find(&system, "name = ?", "System").Error; err != nil {
+		return err
+	} else {
```
yashmehrotra, line 135: else not really required here

db/init.go
```diff
@@ -82,17 +101,42 @@ func Init(connection string) error {
...
+		api.SystemUserID = &system.ID
+		logger.Infof("System user ID: %s", system.ID.String())
+	}
+	return nil
```
yashmehrotra, line 139:
```suggestion
	system := api.Person{}
	if err = Gorm.Find(&system, "name = ?", "System").Error; err != nil {
		return err
	}
	api.SystemUserID = &system.ID
	logger.Infof("System user ID: %s", system.ID.String())
	return nil
```

db/migrations/004_incidents.sql
moshloop: These are core fields used for incident lifecycle automation, and later for machine learning which means that would be queried alot

PR #172, opened 2022-10-28
chore: update release workflow to update incident-commander-chart

Joeshiett:
@moshloop please do have a look and give feedback thanks.


moshloop:
@Joeshiett this workflow needs to be changed to match the others, i.e. updat eethe incident-commander-chart with the new incident commander version of the container , it shouldn't be publishing any charts from this repo

PR #188, opened 2022-11-22
feat: invite user api

yashmehrotra:


auth/controllers.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	link, err := k.createRecoveryLink(identity.Id)
+	if err != nil {
+		errMsg := []byte(fmt.Sprintf(`{"error": "%v", "message": "error creating recovery link"}`, err))
+		return c.JSONBlob(http.StatusInternalServerError, errMsg)
+	}
+
+	body := fmt.Sprintf(inviteUserTemplate, reqData.FirstName, link)
+	mail := utils.NewMail("team@flanksource.com", reqData.Email, "User Invite", body, "text/html")
```
moshloop, line 36: Can we make this a command line argument ?

yashmehrotra: The `from` part ?

auth/controllers.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	link, err := k.createRecoveryLink(identity.Id)
+	if err != nil {
+		errMsg := []byte(fmt.Sprintf(`{"error": "%v", "message": "error creating recovery link"}`, err))
+		return c.JSONBlob(http.StatusInternalServerError, errMsg)
+	}
+
+	body := fmt.Sprintf(inviteUserTemplate, reqData.FirstName, link)
+	mail := utils.NewMail(reqData.Email, "User Invite", body, "text/html")
```
moshloop, line 36: Can we use a mail package? i.e. mail.New()

yashmehrotra: Yeah, makes more sense. Also created a new type so we can do:
```golang
m := mail.New(args)
err := m.Send()
```

mail/mailer.go
```diff
@@ -0,0 +1,32 @@
...
+	m := gomail.NewMessage()
+	m.SetHeader("From", FromAddress)
+	m.SetHeader("To", to)
+	m.SetHeader("Subject", subject)
+	m.SetBody(contentType, body)
+	return &Mail{message: m}
+}
+
+func (m Mail) Send() error {
+	host := os.Getenv("SMTP_HOST")
```
moshloop, line 26: Can we map these through on n the helm charts ?

PR #196, opened 2022-12-09
feat: evidence script evaluation using cel

yashmehrotra:


jobs/evidence_script_cel.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	prg, err := env.Program(ast)
+	if err != nil {
+		return "", err
+	}
+
+	out, _, err := prg.Eval(map[string]any{
+		"config":    evidence.Config.Config,
+		"component": evidence.Component.AsMap(),
```
moshloop, line 61: We should probably have incident here as well

jobs/evidence_script_cel.go
```diff
@@ -0,0 +1,68 @@
...
+func evaluate(evidence api.Evidence) (string, error) {
+	env, err := cel.NewEnv(
+		cel.Variable("config", cel.AnyType),
+		cel.Variable("component", cel.AnyType),
+	)
+	if err != nil {
+		return "", err
+	}
+
+	ast, iss := env.Compile(evidence.Script)
```
moshloop, line 49: I think this compiled scripted should be cached

db/evidences.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/flanksource/incident-commander/api"
+	"github.com/google/uuid"
+
+	"github.com/flanksource/commons/logger"
+)
+
+func GetEvidenceScripts() []api.Evidence {
+	var evidences []api.Evidence
+	openIncidentsSubQuery := Gorm.Table("hypotheses").Select("id").Where("incident_id IN (?)",
+		Gorm.Table("incidents").Select("id").Where("closed IS NULL AND resolved IS NULL"))
```
moshloop, line 13:
```suggestion
		Gorm.Table("incidents").Select("id").Where("closed IS NULL"))
```

A resolved incident could be re-opened if the script starts failing

db/evidences.go
```diff
@@ -0,0 +1,32 @@
...
+	err := Gorm.Table("evidences").
+		Joins("Config").Joins("Component").
+		Where("evidences.hypothesis_id IN (?)", openIncidentsSubQuery).
+		Where("evidences.script IS NOT NULL").
+		Where("definition_of_done != true").
```
moshloop, line 18: I think this should be other way around - only definition of done evidence items can have scripts attached

jobs/evidence_script_cel.go
```diff
@@ -0,0 +1,68 @@
...
+
+		var result string
+		done, err := strconv.ParseBool(output)
+		if err != nil {
+			result = "Script should evaluate to a boolean value"
+		}
+		if err = db.UpdateEvidenceScriptResult(evidence.ID, done, result); err != nil {
+			logger.Errorf("Error persisting evidence script result: %v", err)
+		}
+	}
```
moshloop, line 37: We probably want to track the evidence items that have changed status, and then update the status accordingly:

- If all evidence items marked as the definition of done are true && the status != resolved/closed, then update the status to resolved
- If any evidence items marked as the definition of done are false && the status == resolved, then update status to the status prior to resolved e.g. investigating

jobs/jobs.go
```diff
@@ -6,21 +6,28 @@ import (
 )
 
 const (
-	TeamComponentOwnershipSchedule = "@every 15m"
+	TeamComponentOwnershipSchedule  = "@every 15m"
+	EvaluateEvidenceScriptsSchedule = "@every 30m"
```
moshloop, line 10:
```suggestion
	EvaluateDefinitionofDone = "@every 5m"
```

api/evidence.go
```diff
@@ -13,6 +13,8 @@ type Evidence struct {
 	ConfigID         *uuid.UUID    `json:"config_id"`
 	ConfigChangeID   *uuid.UUID    `json:"config_change_id"`
 	ConfigAnalysisID *uuid.UUID    `json:"config_analysis_id"`
+	Config           ConfigItem    `json:"config,omitempty" gorm:"foreignKey:ConfigID;references:ID"`
```
moshloop, line 16: Config and Component are optional, and I don't think we want to embed them?

yashmehrotra: type xyz struct {
Evidence
Config
Incident
Compn
}

jobs/evidence_script_cel.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/flanksource/commons/logger"
+	"github.com/google/cel-go/cel"
+	"github.com/google/uuid"
+
+	"github.com/flanksource/incident-commander/db"
+)
+
+func EvaluateEvidenceScripts() {
+	logger.Debugf("Evaluating evidence scripts")
+
```
moshloop, line 16:
```suggestion
```

jobs/jobs.go
```diff
@@ -1,26 +1,41 @@
...
 var FuncScheduler = cron.New()
 
-func ScheduleFunc(schedule string, fn func()) (interface{}, error) {
+func ScheduleFunc(schedule string, fn func()) (any, error) {
 	return FuncScheduler.AddFunc(schedule, fn)
 }
 
 func Start() {
-	// running first at startup and then with the schedule
+	prgCache = cache.New(24*time.Hour, 1*time.Hour)
```
moshloop, line 26: I don't think this should be in this file, rather in the file where the cache is used. 
Alternatively in a cel/cache.go which is used by a cel.Evaluate(script, args) which becomes reusable for other use cases.

db/evidences.go
```diff
@@ -0,0 +1,47 @@
...
+
+type EvidenceScriptInput struct {
+	api.Evidence
+	ConfigItem api.ConfigItem `json:"config,omitempty" gorm:"foreignKey:ConfigID;references:ID"`
+	Component  api.Component  `json:"component,omitempty"`
+	Hypothesis Hypothesis
+}
+
+func GetEvidenceScripts() []EvidenceScriptInput {
+
```
moshloop, line 23:
```suggestion
```

db/incidents.go
```diff
@@ -0,0 +1,35 @@
...
+            FROM evidences
+            INNER JOIN hypotheses ON hypotheses.id = evidences.hypothesis_id
+            INNER JOIN incidents ON incidents.id = hypotheses.incident_id
+            WHERE
+                incidents.status != 'closed' AND
+                evidences.definition_of_done = true AND
+                incidents.id IN (?)
+            GROUP BY incidents.id
+        )
+        UPDATE incidents
```
moshloop, line 21: We probably need a trigger here to add the event = Or alternatively we just query the incidents that can change, and then update them individual + insert history item

jobs/jobs.go
yashmehrotra: For now, I'll just put this in the same file. I originally went with the `cel` package approach then realised it is too specific to evidence's scope now, so will do a refactor when we work on similar functionality again

db/migrations/004_incidents.sql
```diff
@@ -195,6 +197,22 @@ CREATE TRIGGER incident_to_incident_history
     FOR EACH ROW
     EXECUTE PROCEDURE insert_incident_created_in_incident_history();
 
+-- Insert incident status updates in incident_histories
+CREATE OR REPLACE FUNCTION insert_incident_status_update_in_incident_history()
+RETURNS TRIGGER AS $$
+BEGIN
+    INSERT INTO incident_histories(incident_id, created_by, type, description) VALUES (NEW.id, NEW.created_by, 'incident.status_updated', NEW.status);
```
moshloop, line 204: I don't think NEW.created_by is the correct value - 

Long term we want to link the oauth tokens generated by kratos proxy into https://postgrest.org/en/v10.0/auth.html#web-users-sharing-role so that we can get the identity of the user logged in, so maybe we should create a get_current_user() function and In then in the job set a local session variable = System, and have the function look at the postgrest session variable and then fallback to System user

Issue #194, opened 2022-12-07
Add responder column to incident_history table


mainawycliffe:
In order to be able to construct an incident changelog/history, we need important information linking to the actual item that was affected by the history; for instance, `responder.added` should link to the added responder as an incident can have multiple responders. 

Other relations possibly missing:

- [ ] Evidences
- [ ] Configs
- [ ] Config Analysis

PR #199, opened 2022-12-14
feat: add foreign key relationships for incident histories

yashmehrotra:
Fixes: #194


db/migrations/107_incident_history_relationships.sql
```diff
@@ -0,0 +1,13 @@
+-- +goose Up
```
moshloop, line 1: I think the the number 99post sql needs to be updated as well.

Issue #197, opened 2022-12-13
Add default admin user


moshloop:
With user/pass = admin/admin

PR #200, opened 2022-12-14
feat: add default admin user with password

yashmehrotra:


yashmehrotra:
Fixes: #197

auth/users.go
```diff
@@ -30,3 +38,49 @@ func (k *KratosHandler) createRecoveryLink(id string) (string, error) {
 	}
 	return resp.GetRecoveryLink(), nil
 }
+
+func (k *KratosHandler) createAdminIdentity() (string, error) {
+	var adminPassword string
+	adminPassword = os.Getenv("ADMIN_PASSWORD")
```
moshloop, line 44: I think we need to wire this through to the helm chart

yashmehrotra: @moshloop Have created the PR for that here: https://github.com/flanksource/incident-commander-chart/pull/15

PR #205, opened 2022-12-21
Snapshot resources

yashmehrotra:


cmd/root.go
```diff
@@ -39,6 +40,7 @@ func ServerFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&publicEndpoint, "public-endpoint", "http://localhost:8080", "Public endpoint that this instance is exposed under")
 	flags.StringVar(&apmHub, "apm-hub", "http://apm-hub:8080", "APM Hub URL")
 	flags.StringVar(&configDb, "config-db", "http://config-db:8080", "Config DB URL")
+	flags.StringVar(&snapshot.CanaryCheckerURL, "canary-checker", "http://canary-checker:8080", "Canary Checker URL")
```
moshloop, line 43: Why can't we just use api.CanaryChckerPath ?

snapshot/controllers.go
```diff
@@ -0,0 +1,27 @@
...
+	"strconv"
+
+	commonUtils "github.com/flanksource/commons/utils"
+	"github.com/labstack/echo/v4"
+)
+
+func Topology(c echo.Context) error {
+	id := c.Param("id")
+	related, _ := strconv.ParseBool(c.QueryParam("related"))
+	directory := commonUtils.RandomString(12)
```
moshloop, line 16: I would use a snapshot-timestamp instead

cmd/root.go
yashmehrotra: By bad, I didn't see that flag at first, will be using that now

snapshot/controllers.go
yashmehrotra: 👍🏽

snapshot/dump.go
```diff
@@ -0,0 +1,199 @@
...
+		Total   int               `json:"total"`
+		Results []json.RawMessage `json:"results"`
+	}
+
+	for _, row := range rows {
+		payload := payloadBody{
+			ID:   row.ExternalID,
+			Type: row.Type,
+			// TODO: Yash - Change start value
+			Start: "15m",
```
yashmehrotra, line 167: @moshloop What value should we keep here ?

snapshot/snapshot.go
```diff
@@ -0,0 +1,188 @@
...
+		if err != nil {
+			return err
+		}
+		resources.merge(relatedResources)
+	}
+
+	return resources.dump(directory)
+}
+
+func fetchRelatedIDsForComponent(componentIDs []string) (resource, error) {
```
moshloop, line 92: Can we 

a) Move most of the SQL into functions like https://github.com/flanksource/incident-commander/pull/208/files#diff-1a2d4a2b04d8d47065438c2b9deee4a37df131c4b8dcf5076db38f42abbe4da1R13 e.g. `lookup_related_components(id, depth=0)`
Depth being defaulted to 0, and when doing > 0 doing the recursive call to get child components

b) Abstract the sql function calls in the db package e.g. LookupRelatedComponents(), LookupIncidentsByComponent()

utils/http.go
```diff
@@ -0,0 +1,39 @@
+package utils
+
+import (
+	"bytes"
+	"io"
+	"net/http"
+)
+
+func HTTPGet(url string) (string, error) {
```
moshloop, line 9: Rather use: https://pkg.go.dev/github.com/flanksource/commons@v1.6.1/http

utils/zip.go
```diff
@@ -0,0 +1,46 @@
+package utils
+
+import (
+	"archive/zip"
+	"io"
+	"os"
+	"path/filepath"
+)
+
+func Zip(src, dst string) error {
```
moshloop, line 10: Move to https://pkg.go.dev/github.com/flanksource/commons@v1.6.1/files

snapshot/dump.go
```diff
@@ -0,0 +1,199 @@
...
+
+	err = dumpTable("config_analysis", "config_id", "?", configIDs, directory)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func dumpLogs(directory string, componentIDs []string) error {
```
moshloop, line 140: Can we extract a GetLogsByComponent function + expose it as an API? This can then use LookupRelatedComponents(), we would then want to add type, name and external_id as tags onto the resulting log messages if lookupId != componentId. 

i.e. If I am using the API and select logs for a pod, it will return logs directly without any type,name,external_id tags as they would all be same - But if I select a Deployment I am going to get logs from multiple pods and then need to return tags with thoses values.

snapshot/dump.go
```diff
@@ -0,0 +1,199 @@
...
+	return writeToCSVFile(csvDirectory, table+".csv", columnNames, rows)
+}
+
+func dumpComponents(directory string, componentIDs []string) error {
+	if len(componentIDs) == 0 {
+		return nil
+	}
+	var jsonBlobs []map[string]any
+
+	endpoint, err := url.JoinPath(api.CanaryCheckerPath, "/api/topology")
```
moshloop, line 52: Rather use the sdk for this: https://github.com/flanksource/incident-commander/blob/main/rules/job.go#L19

snapshot/controllers.go
```diff
@@ -0,0 +1,59 @@
...
+	if err := os.MkdirAll(directory, os.ModePerm); err != nil {
+		errMsg := []byte(fmt.Sprintf(`{"error": "%v", "message": "error creating directory"}`, err))
+		return c.JSONBlob(http.StatusInternalServerError, errMsg)
+	}
+	if err := topologySnapshot(id, related, directory); err != nil {
+		errMsg := []byte(fmt.Sprintf(`{"error": "%v", "message": "error snapshotting topology"}`, err))
+		return c.JSONBlob(http.StatusInternalServerError, errMsg)
+	}
+
+	return c.File(directory + ".zip")
```
moshloop, line 26: Do we not need to cleanup these directories/files

components/logs.go
```diff
@@ -0,0 +1,65 @@
...
+	"net/url"
+	"strings"
+
+	"github.com/flanksource/commons/http"
+
+	"github.com/flanksource/incident-commander/api"
+	"github.com/flanksource/incident-commander/db"
+)
+
+func GetLogsByComponent(componentID, logStart string) (api.LogsResponse, error) {
```
moshloop, line 15: This should be start and end, even though we have implemented the end functionality yet

snapshot/dump.go
```diff
@@ -0,0 +1,158 @@
...
+		if logs.Total == 0 {
+			continue
+		}
+
+		jsonLogs, err := json.Marshal(logs.Results)
+		if err != nil {
+			return err
+		}
+
+		err = writeToLogFile(ctx.Directory, componentID+".log", jsonLogs)
```
moshloop, line 151: I think we want to use `{type}-{name}-{short hash of componentid}.log`

moshloop (changes requested):

snapshot/controllers.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/flanksource/incident-commander/api"
+	"github.com/labstack/echo/v4"
+)
+
+func Topology(c echo.Context) error {
+	id := c.Param("id")
+	related, _ := strconv.ParseBool(c.QueryParam("related"))
+	logStart := c.QueryParam("logStart")
+	logEnd := c.QueryParam("logEnd")
```
moshloop, line 18: Can we rename these to just start and end to be consistent, as well as in the function args ?

snapshot/controllers.go
```diff
@@ -0,0 +1,92 @@
...
+		return c.JSON(http.StatusInternalServerError, api.HTTPErrorMessage{
+			Error:   err.Error(),
+			Message: "Error creating directory",
+		})
+	}
+
+	ctx := SnapshotContext{
+		Directory: directory,
+		LogStart:  logStart,
+		LogEnd:    logEnd,
```
moshloop, line 33: nit: this entire block could be refactored into NewSanpshotContext(c) which then uses echo.Bind

PR #219, opened 2023-01-12
feat: job history for responder jobs

yashmehrotra:


responder/comments_sync.go
```diff
@@ -42,22 +43,28 @@ func SyncComments() {
 	}
 
 	for _, responder := range responders {
+		jobHistory := models.NewJobHistory("ResponderCommentSync", "responder", responder.ID.String())
```
moshloop, line 46: I think we should probably only be creating a single history event per unique client

yashmehrotra: It would be unique per responder because a responder is only mapped to one external client

An incident can be mapped to multiple clients by mapping multiple responders to the incident respectively

moshloop: Yes, but if I have 1000 incidents with 1 responder per incident I still get 1000 jobs every few minutes. 

I should get 1 job, with a success count of 1000 -  For now we can just add a global sync job, and refeactor at a later point

PR #226, opened 2023-01-24
fix: align secret behaviour with canary-checker chart

BrendanGalloway:


chart/values.yaml
```diff
@@ -12,10 +12,9 @@ image:
 disablePostgrest: false
 
 db:
-  enabled: true
+  create: true
   secretKeyRef:
-    create: true
-    name: postgres-connection
+    name: incident-commander-postgres
```
moshloop, line 17: Why are we renaming this?

PR #229, opened 2023-01-25
feat: inject jwt token for postgrest proxy

yashmehrotra:
Merge with: https://github.com/flanksource/incident-commander-chart/pull/24

### DIFF
--- a/auth/middleware.go
+++ b/auth/middleware.go
@@ -6,26 +6,28 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/flanksource/commons/logger"
 	"github.com/golang-jwt/jwt/v4"
 	"github.com/labstack/echo/v4"
 	client "github.com/ory/client-go"
+	"github.com/patrickmn/go-cache"
 )
 
 const DefaultPostgrestRole = "postgrest_api"
 
-var postgrestJWTToken string
-
 type kratosMiddleware struct {
-	client    *client.APIClient
-	jwtSecret string
+	client     *client.APIClient
+	jwtSecret  string
+	tokenCache *cache.Cache
 }
 
 func (k *KratosHandler) KratosMiddleware() *kratosMiddleware {
 	return &kratosMiddleware{
-		client:    k.client,
-		jwtSecret: k.jwtSecret,
+		client:     k.client,
+		jwtSecret:  k.jwtSecret,
+		tokenCache: cache.New(3*24*time.Hour, 12*time.Hour),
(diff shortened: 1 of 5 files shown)


auth/middleware.go
```diff
@@ -35,6 +36,14 @@ func (k *kratosMiddleware) Session(next echo.HandlerFunc) echo.HandlerFunc {
 		if !*session.Active {
 			return c.String(http.StatusUnauthorized, "Unauthorized")
 		}
+
+		// Adding Authorization Token for PostgREST
+		token, err := k.generateDBToken(session.Identity.GetId())
+		if err != nil {
+			logger.Errorf("Error generating JWT Token: %v", err)
+		}
+		c.Request().Header.Add("Authorization", fmt.Sprintf("Bearer %s", token))
```
moshloop, line 45: We might want to cache the JWT token and generating a new token for every db request will add a lot of latency and CPU

auth/middleware.go
```diff
@@ -35,9 +38,20 @@ func (k *kratosMiddleware) Session(next echo.HandlerFunc) echo.HandlerFunc {
 		if !*session.Active {
 			return c.String(http.StatusUnauthorized, "Unauthorized")
 		}
+
+		if postgrestJWTToken == "" {
```
moshloop, line 42: I think this will need to be cached by identity id, with an expiry that matches the token expiry

auth/middleware.go
```diff
@@ -59,3 +72,25 @@ func (k *kratosMiddleware) validateSession(r *http.Request) (*client.Session, er
 	}
 	return session, nil
 }
+
+func (k *kratosMiddleware) generateDBToken(id string) (string, error) {
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, jwt.MapClaims{
+		"role": DefaultPostgrestRole,
```
moshloop, line 78: I think we need an expiry and then use that same expiry in the cache:
```
		// A usual scenario is to set the expiration time relative to the current time
		ExpiresAt: jwt.NewNumericDate(time.Now().Add(24 * time.Hour)),
```

yashmehrotra: @moshloop Since the cache key works on sessionID, and users never have access to the token, we don't have to worry about expiring the token.

Postgrest in itself supports expiry by using `exp: unix epoch time`, but if we do that, we have to check everytime if a token is expired or not.

The current approach just uses sessionID which is always validated first, and when the session expires a new token is created for that session+user  and the cache's default expiry will cleanup the old token

PR #237, opened 2023-02-08
fix: use postgrest_api when auth is disabled

moshloop:


cmd/root.go
```diff
@@ -46,6 +46,7 @@ func ServerFlags(flags *pflag.FlagSet) {
 	flags.DurationVar(&rules.Period, "rules-period", 5*time.Minute, "Period to run the rules")
 	flags.BoolVar(&disablePostgrest, "disable-postgrest", false, "Disable postgrest")
 	flags.StringVar(&mail.FromAddress, "email-from-address", "no-reply@flanksource.com", "Email address of the sender")
+	flags.StringVar(&db.PostgresDBAnonRole, "postgrest-anon-role", "postgres_anon", "Postgres anonymous role")
```
adityathebe, line 49: Just for backward compatibility, I think the default value should be `postgrest_anon` since I assume most people have that user setup already

yashmehrotra: Yeah, else we have to make changes in duty and the helm charts

moshloop, line 49:
```suggestion
	flags.StringVar(&db.PostgresDBAnonRole, "postgrest-anon-role", "postgrest_anon", "Postgres anonymous role")
```

PR #241, opened 2023-02-13
feat: Custom renderer

### LINKED ISSUE #235: Custom renders per component / property type
i.e. Using a https://github.com/jvatic/goja-babel provide an option to specify a JSX renderer content at a SystemTemplate level.

e.g. 

```yaml
components: []
renderers: 
    components:
        - name: abc
          type: EKSCluster
          jsx: "..."

     properties:
       - name: cpu
          jsx: "..."
```       

This would generate components with the name: `<System template name>-{component|property}-{name|kind}-name

adityathebe:
Resolves: https://github.com/flanksource/incident-commander/issues/235


topology/controllers.go
```diff
@@ -0,0 +1,111 @@
...
+	JS   string
+}
+
+// GetCustomRenderer returns an application/javascript HTTP response
+// with custom components and a registry.
+// This registry needs to be used to select custom components
+// for rendering of properties and cards.
+func GetCustomRenderer(ctx echo.Context) error {
+	// 1. Read the template of the topology
+	sysTplID := ctx.QueryParams().Get("id")
```
moshloop, line 38: I think we will need to maintain a cache as compilation will be expensive

moshloop, line 38: from a code style perspective, comments should say why, not what. Variable names should not be abbreviated except for the few common e.g. id
```suggestion
	id := ctx.QueryParams().Get("id")
```

topology/controllers.go
```diff
@@ -0,0 +1,111 @@
...
+{{range $k, $v := .}}
+const {{$k}} = {{$v.JS}}
+{{end}}
+const componentRegistry = {
+	{{range $k, $v := .}}"{{$k}}": {{$k}},
+	{{end}}
+};
+`
+
+func errorResonse(c echo.Context, err error, code int) error {
```
moshloop, line 108:
```suggestion
func errorResponse(c echo.Context, err error, code int) error {
```

topology/controllers.go
```diff
@@ -0,0 +1,111 @@
...
+// for rendering of properties and cards.
+func GetCustomRenderer(ctx echo.Context) error {
+	// 1. Read the template of the topology
+	sysTplID := ctx.QueryParams().Get("id")
+	results, err := QueryRenderComponents(ctx.Request().Context(), sysTplID)
+	if err != nil {
+		return errorResonse(ctx, err, http.StatusBadRequest)
+	}
+
+	// 2. Create a registry of all the components
```
yashmehrotra, line 44: Don't use numbering for commments

topology/controllers.go
```diff
@@ -0,0 +1,111 @@
...
+	var components = make(map[string]component)
+	for _, r := range results {
+		if err := compileComponents(components, r.Components, false); err != nil {
+			return errorResonse(ctx, err, http.StatusInternalServerError)
```
yashmehrotra, line 48: You can use
 ```go
		return c.JSON(http.StatusInternalServerError, api.HTTPErrorMessage{
			Error:   err.Error(),
			Message: "Error message for user to see",
		})
```

or wrap this in your errorResponse function, but prefer using api.HTTPErrorMessage

File: topology/controllers.go [hunk 5]
```diff
@@ -0,0 +1,111 @@
...
+	return ctx.Stream(http.StatusOK, "application/javascript", registryResp)
+}
+
+func compileComponents(output map[string]component, components []api.RenderComponent, isProp bool) error {
+	if err := babel.Init(len(components)); err != nil {
+		return fmt.Errorf("failed to init babel; %w", err)
+	}
+
+	for _, c := range components {
+		res, err := babel.TransformString(c.JSX, map[string]interface{}{
```
yashmehrotra, line 70:
```suggestion
		res, err := babel.TransformString(c.JSX, map[string]any{
```

topology/controllers.go
```diff
@@ -0,0 +1,111 @@
...
+			"plugins": []string{
+				"transform-react-jsx",
+				"transform-block-scoping",
+			},
+		})
+		if err != nil {
+			return fmt.Errorf("error transforming jsx; %w", err)
```
yashmehrotra, line 77: Use `:` instead of `;`

topology/controllers.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+	}
+
+	return nil
+}
+
+func renderComponents(components map[string]component) (io.Reader, error) {
+	var buf bytes.Buffer
+	if err := jsComponentTpl.Execute(&buf, components); err != nil {
+		return nil, fmt.Errorf("error generating components; %w", err)
```
yashmehrotra, line 92:
```suggestion
		return nil, fmt.Errorf("error generating components: %w", err)
```

topology/query.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/flanksource/incident-commander/api"
+	"github.com/flanksource/incident-commander/db"
+)
+
+func QueryRenderComponents(ctx context.Context, systemTemplateID string) ([]api.Renderers, error) {
```
yashmehrotra, line 12: @moshloop Should we keep this in `db/topology.go` as we keep all the database interactions in the db package

topology/query.go
```diff
@@ -0,0 +1,34 @@
...
+	rows, err := db.Gorm.WithContext(ctx).Table("templates").Select("spec->'renderers'").Where("id = ?", systemTemplateID).Rows()
+	if err != nil {
+		return nil, fmt.Errorf("failed to query renderers(); %w", err)
```
yashmehrotra, line 15:
```suggestion
		return nil, fmt.Errorf("failed to query renderers: %w", err)
```

topology/query.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+	defer rows.Close()
+
+	var results []api.Renderers
+	for rows.Next() {
+		var renderer api.Renderers
+		var s string
+		if err := rows.Scan(&s); err != nil {
+			return nil, fmt.Errorf("rows.Scan(); %w", err)
```
yashmehrotra, line 24: Use descriptive error message
```suggestion
			return nil, fmt.Errorf("error scanning row: %w", err)
```

topology/query.go
```diff
@@ -0,0 +1,34 @@
...
+		}
+
+		if err := json.Unmarshal([]byte(s), &renderer); err != nil {
+			return nil, fmt.Errorf("json.Unmarshal(); %w", err)
```
yashmehrotra, line 28: same as above

topology/controllers.go
adityathebe: I agree.
 
 It's just a convention I'm used to following where I write the steps before I start writing the code.
 Will remove it.

topology/query.go
adityathebe: To me they're both equally descriptive. I'll make the change if that's the convention here.

topology/query.go
adityathebe: I assumed that was the convention I am supposed to follow. I took `pkg/snapshot` as a reference but looks like `pkg/rules` also does it.

topology/controllers.go
adityathebe: Makes sense.

Added an inmem cache.

PR #245, opened 2023-02-17
feat: Pushmode

adityathebe:
TODO

- [x] implement basic auth
- [x] create a dummy template for the upstream server
- [x] inject additional labels from the flags
- [ ] ~~handle events in batches~~


jobs/jobs.go
```diff
@@ -12,6 +15,7 @@ const (
 	ResponderCommentsSyncSchedule   = "@every 1h"
 	ResponderConfigSyncSchedule     = "@every 1h"
 	CleanupJobHistoryTableSchedule  = "@every 24h"
+	upstreamPushSchedule            = "@every 5s"
```
adityathebe, line 18: @moshloop This is 5s only for testing purpose. But what should be the final value? Maybe this could also be a flag ... ?

jobs/jobs.go
```diff
@@ -48,5 +56,12 @@ func Start() {
 		logger.Errorf("Failed to schedule job for cleaning up job history table: %v", err)
 	}
 
+	if config.UpstreamConf.IsFilled() {
+		refTime := time.Now().Add(time.Minute * -20)
```
adityathebe, line 60: @moshloop refTime will eventually come from the database. Will need to create a new table to store all the upstream push times.

jobs/jobs.go
moshloop: This shouldn't be a schedule, but a loop - the changelog table acts as an event queue: https://github.com/flanksource/incident-commander/blob/main/events/events.go#L67

db/models/changelog.go
```diff
@@ -0,0 +1,14 @@
+package models
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type Changelog struct {
```
moshloop, line 9: Move this to duty with corresponding schemas - Also rename to PushQueue

api/upstream.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"strings"
+	"time"
+
+	"github.com/flanksource/duty/models"
+)
+
+// ConfigChanges consists of data about changes to
+// components, configs, analysis.
+type ConfigChanges struct {
```
moshloop, line 12:
```suggestion
type PushData struct {
```

cmd/root.go
```diff
@@ -48,6 +49,11 @@ func ServerFlags(flags *pflag.FlagSet) {
 	flags.StringVar(&mail.FromAddress, "email-from-address", "no-reply@flanksource.com", "Email address of the sender")
 	flags.StringVar(&db.PostgresDBAnonRole, "postgrest-anon-role", "postgrest_anon", "Postgres anonymous role")
 
+	// Flags for upstream push
+	flags.StringVarP(&upstreamConfig.URL, "upstream", "u", "https://incident-commander.canary.lab.flanksource.com", "central incident commander instance to push configs to")
```
moshloop, line 53:
```suggestion
	flags.StringVar(&upstreamConfig.URL, "upstream-url", "", "central incident commander instance to push configs to")
```

go.mod
```diff
@@ -22,7 +22,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	gopkg.in/gomail.v2 v2.0.0-20160411212932-81ebce5c23df
 	gorm.io/gorm v1.24.5
-	k8s.io/apimachinery v0.24.4
+	k8s.io/apimachinery v0.25.1
```
moshloop, line 25: Why are we bumping this ?

db/models/changelog.go
```diff
@@ -0,0 +1,14 @@
...
+	ID        uuid.UUID
+	ItemID    uuid.UUID `gorm:"column:item_id"`
+	Timestamp time.Time `gorm:"column:tstamp"`
```
moshloop, line 12:
```suggestion
	CreatedAt time.Time `gorm:"column:created_at"`
```

db/models/changelog.go
```diff
@@ -0,0 +1,14 @@
...
+	TableName string    `gorm:"column:tablename"`
```
moshloop, line 13:
```suggestion
	Table string    `gorm:"column:table"`
```

go.mod
adityathebe: Not sure, how this got bumped. Will revert it.

events/events.go
```diff
@@ -17,112 +17,125 @@ import (
...
+	if err != nil {
+		logger.Errorf("error deleting event from event_queue table with id:%s", event.ID.String())
+		return tx.Rollback().Error
 	}
+	return tx.Commit().Error
+}
 
-	// Keep on iterating till the queue is empty
+// consumeEventsUntilEmpty consumes events forever until the event queue is empty.
+func consumeEventsUntilEmpty(ctx context.Context, config Config) {
```
yashmehrotra, line 130: The name is misleading, it should indicate that it keeps on polling events in  an infinite for loop

events/upstream_push.go
```diff
@@ -0,0 +1,138 @@
...
+		case "component_relationships":
+			output[tableName] = append(output[tableName], []string{cl.Properties["component_id"], cl.Properties["relationship_id"], cl.Properties["selector_id"]})
+		case "config_component_relationships":
+			output[tableName] = append(output[tableName], []string{cl.Properties["component_id"], cl.Properties["config_id"]})
+		case "config_relationships":
+			output[tableName] = append(output[tableName], []string{cl.Properties["related_id"], cl.Properties["config_id"], cl.Properties["selector_id"]})
+		case "check_statuses":
+			output[tableName] = append(output[tableName], []string{cl.Properties["check_id"], cl.Properties["time"]})
+		default:
+			output[tableName] = append(output[tableName], cl.Properties["id"])
```
yashmehrotra, line 133: Set a variable for switch case and run `output[tablename] = append(..., var` only once

upstream/controllers.go
```diff
@@ -0,0 +1,103 @@
...
+		return c.JSON(http.StatusBadRequest, api.HTTPErrorMessage{
+			Error:   err.Error(),
+			Message: "invalid json request",
+		})
+	}
+
+	if err := insertUpstreamMsg(c.Request().Context(), &req); err != nil {
+		return c.JSON(http.StatusInternalServerError, api.HTTPErrorMessage{
+			Error:   err.Error(),
+			Message: "something went wrong",
```
yashmehrotra, line 29: Try to be more descriptive

upstream/controllers.go
```diff
@@ -0,0 +1,103 @@
...
+		})
+	}
+
+	logger.Debugf("Checked at %v", req.CheckedAt)
```
yashmehrotra, line 33: remove before merging

upstream/controllers.go
```diff
@@ -0,0 +1,103 @@
...
+	return nil
+}
+
+func insertUpstreamMsg(ctx context.Context, req *api.PushData) error {
+	if len(req.Canaries) > 0 {
+		if err := db.Gorm.Clauses(clause.OnConflict{UpdateAll: true}).Create(req.Canaries).Error; err != nil {
```
yashmehrotra, line 39: If processing one of this fails, none of the other fields would be processed. I don't think thats a desirable flow

yashmehrotra (changes requested):

events/events.go
adityathebe: I think it's fairly descriptive - the loop is not really infinite because it terminates when the queue is empty.  And it's not necessary to mention how the queue is fully consumed; whether it's with infinite for loop or something else.

Do you have any suggestion?

events/upstream_push.go
adityathebe: Can you please elaborate on this?

events/events.go
yashmehrotra: Sorry my bad, I confused it with the Listen events func

events/upstream_push.go
yashmehrotra:
```golang
		switch tableName {
		case "component_relationships":
			logs = []string{cl.Properties["component_id"], cl.Properties["relationship_id"], cl.Properties["selector_id"]})
		case "config_component_relationships":
			logs = []string{cl.Properties["component_id"], cl.Properties["config_id"]})
		case "config_relationships":
        ...
		default:
			logs = cl.Properties["id"])
    output[tablename] = append(..., logs)
```

events/events.go
```diff
@@ -17,112 +17,125 @@ import (
...
-			// creating dangling connections and to release the locks
-			tx.Commit()
-			return err
-		}
+	selectEventsQuery := fmt.Sprintf(`
+		SELECT * FROM event_queue
+		WHERE attempts <= %d OR ((now() - last_attempt) > '1 hour'::interval)
+		FOR UPDATE SKIP LOCKED
+		LIMIT %d
+	`, eventMaxAttempts, eventQueueBatchSize)
```
yashmehrotra, line 86: Use named args here instead of fmt.Sprint

api/upstream.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+type UpstreamConfig struct {
+	URL      string
+	Username string
+	Password string
+	Labels   []string
+}
+
+func (t *UpstreamConfig) IsFilled() bool {
```
yashmehrotra, line 34: rename to .Valid() bool

api/upstream.go
```diff
@@ -0,0 +1,54 @@
...
+	return t.URL != "" && t.Username != "" && t.Password != ""
+}
+
+func (t *UpstreamConfig) LabelsMap() map[string]string {
+	return sanitizeStringSliceVar(t.Labels)
+}
+
+func sanitizeStringSliceVar(in []string) map[string]string {
```
yashmehrotra, line 42: Can  be moved to utils package

yashmehrotra: @moshloop Should we move this to commons repo ? duty and canary checker also have a similar function ie. parsing raw labels into map[string]string

upstream/controllers.go
```diff
@@ -0,0 +1,103 @@
...
+	"net/http"
+
+	"github.com/flanksource/commons/logger"
+	"github.com/flanksource/incident-commander/api"
+	"github.com/flanksource/incident-commander/db"
+	"github.com/labstack/echo/v4"
+	"gorm.io/gorm/clause"
+)
+
+func UpstreamPushesCtrl(c echo.Context) error {
```
yashmehrotra, line 16: maybe UpstreamPusher ? Since its in the controllers file, we shouldn't have to suffix the `Ctrl` part

adityathebe: It's also part of the `upstream` pkg which is not just for controllers _(I suppose)_. So from the perspective of all the APIs that we expose from the `upstream` pkg I think it's nice to have a suffix for clarity. Unless we just have controllers in the package.

events/events.go
adityathebe: Although it's a dynamically generated query, it's really a static query in essence. The injected variables do not change over the execution of the program & are also in our control. 

I don't think it's necessary to be injected through gorm.

yashmehrotra: We might have to expose the `attempts` and `limit` to the users later on, so have to sanitize the input before sending it to the db

upstream/controllers.go
yashmehrotra: Then its better to use a verb.context method, like `PushUpstream`

upstream/controllers.go
```diff
@@ -0,0 +1,103 @@
...
+			Error:   err.Error(),
+			Message: "something went wrong",
+		})
+	}
+
+	logger.Debugf("Checked at %v", req.CheckedAt)
+	return nil
+}
+
+func insertUpstreamMsg(ctx context.Context, req *api.PushData) error {
```
yashmehrotra, line 37: Move this to db/upstream.go, we keep all the db interactions scoped in that

events/events.go
adityathebe: ah I see

upstream/controllers.go
adityathebe: At this stage, we'll only have one item.

api/upstream.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	for i := range t.Canaries {
+		t.Canaries[i].Labels = mergeMap(t.Canaries[i].Labels, labels)
+	}
+}
+
+// mergeMap will merge map b into a.
+// On key collision, map b takes precedence.
+func mergeMap(a, b map[string]string) map[string]string {
```
adityathebe, line 51: This could go into `commons` as well

api/upstream.go
moshloop: Yeh, makes sense

db/templates.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/flanksource/duty/models"
+	"github.com/google/uuid"
+	"gorm.io/gorm"
+)
+
+// dummyTemplateNamespace is a shared template namespace for
+// downstream instances.
+const dummyTemplateNamespace = "push"
+
+func getDummyTemplate(ctx context.Context, name string) (*models.SystemTemplate, error) {
```
moshloop, line 17: There are not dummy - maybe call them headless templates

api/upstream.go
adityathebe: Added https://github.com/flanksource/commons/pull/58

api/upstream.go
adityathebe: https://github.com/flanksource/commons/pull/58

yashmehrotra (approved):
@moshloop Please have a final look before merging

PR #248, opened 2023-02-21
feat: add state and property update endpoints

yashmehrotra:
Bump duty version after https://github.com/flanksource/duty/pull/39 is merged


auth/controllers.go
```diff
@@ -48,3 +51,52 @@ func (k *KratosHandler) InviteUser(c echo.Context) error {
 	respJSON := []byte(fmt.Sprintf(`{"link": "%s"}`, link))
 	return c.JSONBlob(http.StatusOK, respJSON)
 }
+
+func UpdateAccountState(c echo.Context) error {
```
moshloop, line 55: I think we should create functions for updating that are not echo aware - they just accept params, and then the echo handlers are just parsing the request and calling those functions

PR #249, opened 2023-02-21
updated Ory client to v1.1.17

adityathebe:


adityathebe:
I used this API ref https://github.com/ory/sdk/blob/master/clients/client/go/README.md.

Spun of a new instance of local setup with a fresh DB and the identities were created.

PR #260, opened 2023-03-02
feat: allow basic auth for all endpoints

adityathebe:
Closes https://github.com/flanksource/incident-commander/issues/238
Closes https://github.com/flanksource/incident-commander/issues/257

### DIFF
--- a/auth/middleware.go
+++ b/auth/middleware.go
@@ -73,10 +73,6 @@ func (k *kratosMiddleware) validateSession(r *http.Request) (*client.Session, er
 		return &login.Session, nil
 	}
 
-	if strings.HasPrefix(r.URL.Path, "/upstream_push") && !hasBasicAuth {
-		return nil, errors.New("endpoint requires basic auth")
-	}
-
 	cookie, err := r.Cookie("ory_kratos_session")
 	if err != nil {
 		return nil, err
--- a/auth/middleware.go
+++ b/auth/middleware.go
@@ -1,6 +1,7 @@
 package auth
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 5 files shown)


auth/middleware.go
```diff
@@ -77,6 +73,10 @@ func (k *kratosMiddleware) validateSession(r *http.Request) (*client.Session, er
 		return &login.Session, nil
 	}
 
+	if strings.HasPrefix(r.URL.Path, "/upstream_push") && !hasBasicAuth {
+		return nil, errors.New("endpoint requires basic auth")
+	}
```
adityathebe, line 78: @moshloop should I remove this? So upstream can also work with cookies?

moshloop: Yeh - I don't think we should different auth methods for different endpoints

auth/middleware.go
```diff
@@ -17,17 +18,26 @@ import (
...
+	randString, err := utils.GenerateRandString(30)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate random string: %w", err)
 	}
+
+	return &kratosMiddleware{
+		client:             k.client,
+		jwtSecret:          k.jwtSecret,
+		tokenCache:         cache.New(3*24*time.Hour, 12*time.Hour),
+		authSessionCache:   cache.New(30*time.Minute, time.Hour),
```
adityathebe, line 38: @moshloop I could also potentially use the same cache instance and use custom expiration time for basic auth caches.

auth/middleware.go
```diff
@@ -57,23 +67,31 @@ func (k *kratosMiddleware) validateSession(r *http.Request) (*client.Session, er
...
 
-		username, password, ok := r.BasicAuth()
-		if !ok {
-			return nil, errors.New("endpoint requires basic auth")
-		}
-
 		updateLoginFlowBody := client.UpdateLoginFlowBody{UpdateLoginFlowWithPasswordMethod: client.NewUpdateLoginFlowWithPasswordMethod(username, "password", password)}
 		login, _, err := k.client.FrontendApi.UpdateLoginFlow(r.Context()).Flow(loginFlow.Id).UpdateLoginFlowBody(updateLoginFlowBody).Execute()
 		if err != nil {
 			return nil, fmt.Errorf("failed to update native login flow: %w", err)
```
adityathebe, line 88: Even on auth failure _(when invalid credential is provided)_, I wanted to cache the response for a slightly shorter period of time _(1 minute?)_ so a faulty downstream would not be able to bombard the central instance.

 However, the error returned doesn't have a reliable way to know that the error occurred due to invalid credentials.

moshloop: I think we can cache login failures for 1m irrespective of reason

adityathebe: okay - I guess that's reasonable.

PR #273, opened 2023-03-13
fix: handle db connection close when listening to postgres notifications

adityathebe:
Resolves https://github.com/flanksource/incident-commander/issues/159


events/events.go
```diff
@@ -50,24 +53,51 @@ func ListenForEvents(ctx context.Context, config Config) {
...
-		logger.Errorf("error listening to database notify: %v", err)
+	var listenToPostgresNotify = func(ctx context.Context, pgNotify chan bool) error {
+		conn, err := db.Pool.Acquire(ctx)
+		if err != nil {
+			return fmt.Errorf("error acquiring database connection: %v", err)
+		}
+		defer conn.Release()
+
+		reconnectAttempts = 0
+		logger.Infof("acquired database connection")
```
moshloop, line 69:
```suggestion
```

events/events.go
```diff
@@ -50,24 +53,51 @@ func ListenForEvents(ctx context.Context, config Config) {
...
-		_, err = conn.Conn().WaitForNotification(ctx)
-		if err != nil {
-			logger.Errorf("error waiting for database notifications: %v", err)
+		if err := listenToPostgresNotify(ctx, pgNotify); err != nil {
+			reconnectAttempts++
+			logger.Errorf("error listening to database notifications: %v", err)
+		}
+
+		if reconnectAttempts > dbReconnectMaxAttempt {
+			logger.Fatalf("failed to reconnect to database after %d attempts", dbReconnectMaxAttempt)
```
moshloop, line 96: Fatalf will stop the process which we never want, 

I think use something https://github.com/sethvargo/go-retry, with an ExponentialBack of 1s with a MaxDuration of  5 minutes. 

We would then only log an error message every 5 minutes.

PR #290, opened 2023-03-30
feat: added /logs endpoint

adityathebe:
Resolves https://github.com/flanksource/canary-checker/issues/569


cmd/server.go
```diff
@@ -105,7 +113,9 @@ var Serve = &cobra.Command{
 		forward(e, "/config", configDb)
 		forward(e, "/canary", api.CanaryCheckerPath)
 		forward(e, "/kratos", kratosAPI)
-		forward(e, "/apm", api.ApmHubPath)
+		forward(e, "/apm", api.ApmHubPath) // Deprecated
+
+		forwardWithMiddlewares(e, "/logs", api.ApmHubPath, logSearchMiddleware)
```
moshloop, line 118: I don't think this a middleware - those are normally generic -

cmd/server.go
```diff
@@ -114,27 +124,139 @@ var Serve = &cobra.Command{
...
+			return fmt.Errorf("failed to parse form: %w", err)
+		}
+
+		componentID, ok := form["id"].(string)
+		if !ok {
+			return next(c) // No component ID, so no need to modify the request
+		}
+
+		selectorName, _ := form["name"].(string)
+		componentID = "01872d9f-29c6-3626-2da5-9851502cd6ba" // Test
```
moshloop, line 154: Need to fix

cmd/server.go
adityathebe: modified to use a http handler instead

logs/handler.go
```diff
@@ -0,0 +1,144 @@
...
+			Message: "failed to get component by the given ID",
+		})
+	}
+
+	selectorName, _ := form["name"].(string)
+	logSelector := getLabelsFromLogSelectors(component.LogSelectors, selectorName)
+	if logSelector == nil {
+		return c.JSON(http.StatusBadRequest, api.HTTPError{
+			Error:   "log selector was not found",
+			Message: fmt.Sprintf("log selector with the name '%s' was not found.", selectorName),
```
moshloop, line 50: Add in the valid names into the error message

logs/handler.go
```diff
@@ -0,0 +1,144 @@
...
+		})
+	}
+
+	templater := ktemplate.StructTemplater{
```
moshloop, line 54: switch to commons/template/StructTemplatar

logs/handler.go
```diff
@@ -0,0 +1,144 @@
...
+		return &selectors[0]
+	}
+
+	for _, selector := range selectors {
+		if selector.Name == name {
+			return &selector
+		}
+	}
+
+	logger.Debugf("could not find log selector with name %s", name)
```
moshloop, line 126: Don't log this

logs/handler.go
```diff
@@ -0,0 +1,144 @@
...
+
+	component, err := duty.GetComponent(req.Context(), db.Gorm, componentID)
+	if err != nil {
+		return c.JSON(http.StatusInternalServerError, api.HTTPError{
+			Error:   err.Error(),
+			Message: "failed to get component by the given ID",
+		})
+	}
+
+	selectorName, _ := form["name"].(string)
```
moshloop, line 45: Make the name optional - if none specified use the first log selector.

adityathebe: @moshloop it's already like that.

Issue #47, opened 2022-05-23
RBAC


moshloop:
Permissions:

Admin:

- User Management
- Team Creation and team component selection

Editor:
- Manage rules for their own team
- Create and edit topology/health/scaper owned by them or their team

Commander:
- Can add new responders
- Can create incidents
- Adds and / manages definition of done
- Closing incident

Responder:
- Can comment/link evidence / add hypothesis
- Resolve / Reopen incident
- Can create incidents, and then becomes the commander of that incident

3rd Party Responder
- View logs / config of items associated with incidences they are an active responder to
- Comment and link evidence
- Add hypothesis / update their own hypothesis

PR #307, opened 2023-04-18
feat: rbac with casbin

yashmehrotra:
Fixes: #47


moshloop:
@yashmehrotra  can we add tests?

Some of what I am expecting are:

/db/identities - needs admin
/db/connections - needs admin, even for read
/db/{topology/canary} - deny read for viewer (there could credentials)
/db/{kratos internal tables} - DENY ALL

moshloop:
@yashmehrotra - I think we are missing the courier tables

PR #316, opened 2023-04-25
Tests for push mode

adityathebe:
Resolves: https://github.com/flanksource/incident-commander/issues/313


cmd/server.go
```diff
@@ -120,7 +120,7 @@ var Serve = &cobra.Command{
 		eventHandlerConfig := events.Config{
 			UpstreamConf: upstreamConfig,
 		}
-		go events.ListenForEvents(context.Background(), eventHandlerConfig)
+		go events.ListenForEvents(context.Background(), db.Gorm, eventHandlerConfig)
```
yashmehrotra, line 123: We're passing db object down to 5 functions, do you think we should create an event handler struct ?

```
eventHandler = EventHandler{db: db.Gorm, ctx: context, config: config}

go eventHandler.ListenForEvents()
````

File: upstream/upstream_test.go [hunk 2]
```diff
@@ -0,0 +1,333 @@
...
+	ginkgo.It("should have transferred all the config relationships", func() {
+		compareEntities(testUpstreamDB, testUpstreamDB, &[]models.ComponentRelationship{})
+	})
+
+	ginkgo.It("should have transferred all the config component relationships", func() {
+		compareEntities(testUpstreamDB, testUpstreamDB, &[]models.ConfigComponentRelationship{})
+	})
+})
+
+func populateMonitoredTables(gormDB *gorm.DB) error {
```
yashmehrotra, line 175: Is there any drawback to calling the populate all tables function from duty ?

upstream/suite_test.go
```diff
@@ -0,0 +1,113 @@
+package upstream
```
yashmehrotra, line 1: Can you move this file to the root ?

I also did something similar with rbac, if we use this in root the db setup can be used by all the tests

cmd/server.go
adityathebe: Sure, I'd prefer that actually.

upstream/upstream_test.go
adityathebe: I don't think there's any. I initially started with duty's dummy for a quick test but then stuck with it as that worked fine.

yashmehrotra: Use this function, takes db instance and loads all the dummy data
```
fixtures/dummy/all.go
9:func PopulateDBWithDummyModels(gormDB *gorm.DB) error {
```

cmd/server.go
yashmehrotra: great 👍🏽

upstream/suite_test.go
adityathebe: Make sense. Although a slight problem would be how the packages (`upstream` & `rbac`) would access the db objects. This is unlike duty where all the tests are written in the same package.

yashmehrotra: We can experiment with importing

PR #345, opened 2023-05-14
feat: auto create agent for upstream push

yashmehrotra:
Depends on: https://github.com/flanksource/duty/pull/154

Fixes: https://github.com/flanksource/incident-commander/issues/341

### DIFF
--- a/upstream/controllers.go
+++ b/upstream/controllers.go
@@ -40,7 +40,7 @@ func PushUpstream(c echo.Context) error {
 				Message: "Error while creating/fetching agent",
 			})
 		}
-		agentID = &agent.ID
+		agentID = agent.ID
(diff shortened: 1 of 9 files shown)


upstream/controllers.go
```diff
@@ -28,17 +31,31 @@ func PushUpstream(c echo.Context) error {
 		return c.JSON(http.StatusBadRequest, api.HTTPError{Error: "cluster_name name is required", Message: "cluster name is required"})
 	}
 
-	headlessTplID, ok := templateIDCache.Get(req.ClusterName)
+	agentID, ok := agentIDCache.Get(req.ClusterName)
```
yashmehrotra, line 34: This would require a unique constraint on name

Also, lets change the name from ClusterName to AgentName ? Since it can work from non-cluster environments too

PR #356, opened 2023-05-18
fix: support done expression for checks

adityathebe:
* cel expression didn't not env var about the check. Now it's added.
* added test for incident done definition using checks.
* fixed bug in previous done definition using config test.

Resolves: https://github.com/flanksource/incident-commander/issues/281

### DIFF
--- a/api/check.go
+++ b/api/check.go
@@ -10,6 +10,8 @@ import (
 type Check struct {
 	ID                 uuid.UUID `json:"id"`
 	LastTransitionTime time.Time `json:"last_transition_time"`
+	LastRuntime        time.Time `json:"last_runtime"`
+	CreatedAt          time.Time `json:"created_at"`
 	Status             string    `json:"status"`
 }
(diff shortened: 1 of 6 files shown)


api/check.go
```diff
@@ -0,0 +1,27 @@
...
+
+// AsMap returns a map[string]any representation of the Check object
+// with some additional fields.
+func (t Check) AsMap() map[string]any {
+	m := make(map[string]any)
+	b, _ := json.Marshal(t)
+	_ = json.Unmarshal(b, &m)
+
+	transitionDuration := time.Since(t.LastTransitionTime)
+	m["transition_duration_sec"] = int(transitionDuration.Seconds())
```
moshloop, line 24: Can we change this to age = time.Now() - last_transition? 

and then the expression becomes `check.status == 'healthy' && age > duration("15m")` 

Can we also add the duration and `created` age in

PR #361, opened 2023-05-22
fix: update params for apm hub log search

yashmehrotra:


logs/handler.go
```diff
@@ -66,19 +67,27 @@ func LogsHandler(c echo.Context) error {
 		})
 	}
 
-	modifiedForm := injectLabelsToForm(logSelector.Labels, form)
-	resp, err := makePostRequest(fmt.Sprintf("%s/%s", api.ApmHubPath, "search"), modifiedForm)
+	apmHubPayload := map[string]any{
+		"id":   reqData.ID,
+		"type": logSelector.Type,
+		// TODO: Should these be intersection or union
+		"labels": collections.MergeMap(logSelector.Labels, reqData.Labels),
```
yashmehrotra, line 74: @moshloop Should we do an intersection of labels here or union ?

logs/handler.go
```diff
@@ -66,19 +67,27 @@ func LogsHandler(c echo.Context) error {
 		})
 	}
 
-	modifiedForm := injectLabelsToForm(logSelector.Labels, form)
-	resp, err := makePostRequest(fmt.Sprintf("%s/%s", api.ApmHubPath, "search"), modifiedForm)
+	apmHubPayload := map[string]any{
+		"id":   reqData.ID,
```
moshloop, line 71: I think id should be component.external_id (then we don't need the kuberetes switch)   and type component.type, not logselector type.

yashmehrotra: Done

PR #369, opened 2023-05-30
feat: shoutrrr responder

adityathebe:
Resolves: https://github.com/flanksource/incident-commander/issues/33


notification/notification.go
```diff
@@ -0,0 +1,8 @@
+package notification
+
+import "github.com/flanksource/incident-commander/api"
+
+type INotifier interface {
```
yashmehrotra, line 5: Just use `Notifier` the `I` prefix is not idiomatic

https://go.dev/doc/effective_go#interface-names

notification/notification_test.go
```diff
@@ -0,0 +1,129 @@
...
+		john = &models.Person{
+			ID:   uuid.New(),
+			Name: "James Bond",
+		}
+		tx := db.Gorm.Create(john)
+		Expect(tx.Error).To(BeNil())
+	})
+
+	ginkgo.It("should create a new component", func() {
+		component = &models.Component{
```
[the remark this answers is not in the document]
adityathebe: I think it's good to keep it independent. Later on if any change on duty could suddenly break tests here. Example: a component name change or an incident title change on the duty dummy fixture would break this test.

notification/shoutrrr.go
```diff
@@ -0,0 +1,167 @@
...
+	for _, service := range t.services {
+		if service.config.Filter != "" {
+			if valid, err := evaluateFilterExpression(service.config.Filter, responder); err != nil {
+				logger.Errorf("error evaluating filter expression for service=%q: %v", service.name, err)
+			} else if !valid {
+				continue
+			}
+		}
+
+		view := map[string]any{
```
moshloop, line 82: shouldn't this be passed through to evalauteFilterExpression

events/events.go
```diff
@@ -189,15 +190,28 @@ func reconcileResponderEvent(tx *gorm.DB, event api.Event) error {
...
 	if err != nil {
 		return err
 	}
 
+	teamSpec, err := _responder.Team.GetSpec()
+	if err != nil {
+		return err
+	}
+
+	shoutrrr, err := notification.NewShoutrrrClient(ctx, teamSpec.Notifications)
```
moshloop, line 203: I would loop over each notification, and then create a client for it, not create the intermediary objects to contain multiple objects and multiple errors. 

I think we would also need to loop over the notifications and then emit new events (`notification.<type>`), 1 per notification (if it filters) so that if 1 notification goes through, it doesn't get retried with another that fails

notification/shoutrrr.go
```diff
@@ -0,0 +1,167 @@
...
+
+		var params *types.Params
+		if service.config.Properties != nil {
+			params = (*types.Params)(&service.config.Properties)
+		}
+
+		sendErrors := service.sender.Send(buff.String(), params)
+		for _, err := range sendErrors {
+			if err != nil {
+				logger.Errorf("error sending message to service=%q: %v", service.name, err)
```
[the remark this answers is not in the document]
adityathebe: Do we currently have retry mechanism in place?

events/events.go
adityathebe: > I would loop over each notification, and then create a client for it, not create the intermediary objects to contain multiple objects and multiple errors.

Our Shoutrrr client internally has a separate service - one for each notification. So, we can handle each notification in a case by case basic just like if we would have multiple shoutrrr clients.

PR #373, opened 2023-05-31
feat: use generic reconciler

yashmehrotra:


hack/boilerplate.go.txt
```diff
@@ -0,0 +1,15 @@
+/*
```
moshloop, line 1: Can we remove this, incident commander is not apache licensed

PR #383, opened 2023-06-05
Feat/incident rules test

adityathebe:
Resolves: https://github.com/flanksource/incident-commander/issues/377


adityathebe:
@moshloop The filters and selectors aren't completely implemented right now
https://github.com/flanksource/incident-commander/blob/main/rules/job.go#L108-L136

**For Filters**: only status is implemented. Severity, category and age are unused.
**For component selectors**: Selector and labels are unused.

Should I implement those in this PR?

rules/job.go
```diff
@@ -58,16 +75,23 @@ outer:
...
 				incident.IncidentRuleID = _rule.ID
-				incident.Status = "open"
+				incident.Status = api.IncidentStatusOpen
+				incident.Severity = "Low" // TODO: make this configurable
 				if incident.Type == "" {
 					incident.Type = api.IncidentTypeAvailability
 				}
 				incident.Title = component.Name + " is " + string(component.Status)
+
+				if _, ok := openIncidentsMap[_rule.ID.String()][incident.Title]; ok {
```
adityathebe, line 90: I am checking duplicates by the incident title. Couldn't find a better alternative.

I think we should probably add component id to the incident as well? 
- That'll be much more robust in determining duplicates specially if the incident title template changes.
- Also, if a rule creates an incident when the component status is "error" and then later when the status changes to "warn", we'll have 2 different incidents. If we have component id on the incident, then we can close the old one before creating a new incident.

rules/job.go
```diff
@@ -58,16 +75,23 @@ outer:
...
+					logger.Infof("Incident %s already exists", incident.Title)
+					continue // this rule already created this incident
+				}
+
 				if err := db.Gorm.Create(&incident).Error; err != nil {
```
adityathebe, line 95: This auto created incident doesn't have a commander or a communicator. The ones created manually have both of them. Is this alright @moshloop ?

rules/job.go
moshloop: Can we not do a lookup on the evidence table ?

rules/job.go
moshloop: Yes

rules/job.go
adityathebe: There could be a lot of components and a lookup for each one of them can potentially be a lot, so I just did a bulk query once.

api/evidence.go
```diff
@@ -30,6 +31,16 @@ type Evidence struct {
 	UpdatedAt        time.Time     `json:"updated_at"`
 }
 
+func (evidence *Evidence) BeforeCreate(tx *gorm.DB) (err error) {
+	if evidence.CreatedBy == uuid.Nil {
+		evidence.CreatedBy = *SystemUserID
+	}
+	if evidence.ID == uuid.Nil {
+		evidence.ID = uuid.New()
+	}
```
yashmehrotra, line 40: Do we need this ? can't we do that via schema ?

adityathebe: We can either do it in here or in the gorm struct tag. I guess struct tag is more idiomatic.

PR #397, opened 2023-06-14
feat: periodic reconcile for push mode

adityathebe:


cmd/server.go
```diff
@@ -133,7 +134,7 @@ var Serve = &cobra.Command{
 		go eventHandler.ListenForEvents()
 
 		mgr, err := kopper.Manager(&kopper.ManagerOptions{
-			MetricsBindAddress: ":8081",
+			MetricsBindAddress: fmt.Sprintf(":%d", kopperPort),
```
moshloop, line 137: @yashmehrotra why do we need this? can we not leave this blank ?

db/upstream.go
```diff
@@ -4,71 +4,189 @@ import (
...
+	if err := Gorm.WithContext(ctx).Not(req.ConfigAnalysis).Find(&upstreamMsg.ConfigAnalysis).Error; err != nil {
+		return nil, fmt.Errorf("error fetching config analyses: %w", err)
+	}
+
+	if err := Gorm.WithContext(ctx).Not(req.ConfigChanges).Find(&upstreamMsg.ConfigChanges).Error; err != nil {
+		return nil, fmt.Errorf("error fetching config changes: %w", err)
+	}
+
+	checkStatusQuery := Gorm.WithContext(ctx)
+	for _, cs := range req.CheckStatuses {
```
moshloop, line 43: This will not perform very well as there will be a huge number of statuses - I think analysis, changes, relationships, can also be skipped - they will retry after the config item and check have reconciled

cmd/server.go
yashmehrotra: The default value for this is `8080` which collides with all our default ports, we can either give a port address or give `"0"` to disable it altogether

If we are disabling it, I suggest lets disable it in kopper

jobs/func_job.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/flanksource/commons/logger"
+	"github.com/robfig/cron/v3"
+)
+
+type funcJob struct {
+	name    string                      // name is just an additional context for the job.
+	timeout time.Duration               // optional timeout for the job
+	fn      func(context.Context) error // the actual job
+	runNow  bool                        // whether to run the job now
```
yashmehrotra, line 16: Add comments to top of fields instead of inline

adityathebe: This is a fairly standard practice.  For short descriptions, I prefer this because I can still get the fields formatted in line.

db/upstream.go
```diff
@@ -4,19 +4,80 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/flanksource/duty/models"
 	"github.com/flanksource/incident-commander/api"
 	"gorm.io/gorm/clause"
 )
 
+// TODO: Is it ncessesary to check for agent_id IS NULL here?
```
adityathebe, line 12: @moshloop

PR #416, opened 2023-06-27
refactor: add priority to event handling and improve query

yashmehrotra:
Depends on: https://github.com/flanksource/duty/pull/186


events/events.go
```diff
@@ -32,6 +31,8 @@ const (
 
 	dbReconnectMaxDuration         = time.Minute * 5
 	dbReconnectBackoffBaseDuration = time.Second
+
+	eventWorkers = 3
```
moshloop, line 35: can we make this a cli arg ?

yashmehrotra: Done 👍🏽

PR #417, opened 2023-06-27
feat: access database connection from the context

adityathebe:
* Refactored & fixed bug in upstream test
* Removed vars from testutils. (_Since the tests are now moved to the events package, the vars can be inside the package itself.)_


api/global.go
```diff
@@ -20,36 +21,56 @@ var ApmHubPath string
...
 
+// type alias because the name "Context" collides with gocontext
+// and embedding both wouldn't have been possible.
+type EchoContext = echo.Context
+
 type Context struct {
+	EchoContext
 	gocontext.Context
-	DB         *gorm.DB
+	db         *gorm.DB
```
adityathebe, line 31: made it private and added `DB()` method instead which will return the db with context wrapped.

cmd/server.go
```diff
@@ -42,129 +43,146 @@ var cacheSuffixes = []string{
...
+		if err := rbac.Init(adminUserID); err != nil {
+			logger.Fatalf("Failed to initialize rbac: %v", err)
 		}
+	}
+
+	if !disablePostgrest {
+		forward(e, "/db", "http://localhost:3000", rbac.Authorization(rbac.ObjectDatabase, "any"))
+	}
+
+	if externalPostgrestUri != "" {
```
moshloop, line 83: I think this needs to be an if else

moshloop:
@yashmehrotra  can you give it a onceover ?

db/job_history.go
```diff
@@ -2,24 +2,25 @@ package db
...
+	"github.com/flanksource/incident-commander/api"
 	"github.com/google/uuid"
 )
 
-func PersistJobHistory(h *models.JobHistory) error {
-	if Gorm == nil {
+func PersistJobHistory(ctx *api.Context, h *models.JobHistory) error {
+	if ctx.DB() == nil {
 		return nil
 	}
```
yashmehrotra, line 12: I think we should return an error here, since we want to persist it but database is not there, not sure why this was `nil` before

adityathebe: Makes sense. Although why do we even have to check for nil in this particular db function?
If db is nil, this func will never be called

auth/controllers.go
```diff
@@ -91,6 +93,8 @@ func UpdateAccountState(c echo.Context) error {
 }
 
 func UpdateAccountProperties(c echo.Context) error {
+	ctx := c.(*api.Context)
```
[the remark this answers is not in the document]
adityathebe: I think type assertion in Go is pretty fast - https://stackoverflow.com/a/28027945. 
It pretty much has no overhead at all.

But, I'm not sure how `context.FromEchoContext` would work without type assertion?

db/job_history.go
yashmehrotra: Can't really, I remember putting this because something was failing

auth/controllers.go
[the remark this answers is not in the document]
yashmehrotra: I wasn't saying it from a speed perspective, more from a safety perspective (what if the typecast fails). But I looked more into it and there does not seem to be a better way 

In future, we can use the repo model maybe and attach DB object to handler structs
```
a := AuthHandler{DB: gorm.DB, ...}
e.Handle("/auth/get", a.AuthGet())
e.Handle("/auth/xyz", a.XYZ())
```

But for now this seems fine 👍🏽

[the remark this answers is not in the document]
adityathebe: > I wasn't saying it from a speed perspective, more from a safety perspective
ah I see. I think we are fine in that regards, the middleware gurantees that.

cmd/server.go
```diff
@@ -42,129 +43,144 @@ var cacheSuffixes = []string{
...
+		if err := rbac.Init(adminUserID); err != nil {
+			logger.Fatalf("Failed to initialize rbac: %v", err)
 		}
+	}
 
-		if externalPostgrestUri != "" {
-			forward(e, "/db", externalPostgrestUri, rbac.Authorization(rbac.ObjectDatabase, "any"))
+	if !disablePostgrest {
+		forward(e, "/db", "http://localhost:3000", rbac.Authorization(rbac.ObjectDatabase, "any"))
+	} else if externalPostgrestUri != "" {
```
moshloop, line 81: This needs to be flipped around, if !disablePostgres && externalPostgresUri != "" else if !disablePostgres

adityathebe: I see, so `externalPostgresUri` take precedence over the `disablePostgres` flag?

yashmehrotra: instead of these flags, can we take the input of `--posrgres-uri` as
```
postgrest_opt = One of (none, local-bin, "uri-to-proxy")
```
and take action based on this ?

moshloop: @yashmehrotra I don't think the multi-valued value is more user friendly as there are different types for the value
@adityathebe disable-postgrest takes precedence, but the way the if statement evaluates:

disable == true && external == "" -> do nothing
disable == true && external != "" -> use external (but it shouldn't because its disabled)
!disable && external != "" -> use localhost (it should be using external)

moshloop: I guess that would work.

adityathebe: > disable == true && external != "" -> use external (but it shouldn't because its disabled)

Ah, I see. So we have a different understanding of the `disablePostgrest` flag. 

I thought the flag simply decided whether we fire up our own postgrest server or not. Either we fireup a postgrest server or use an external one, but there has to be a Postgrest server. Without postgrest, would `incident-commander` even be functional?

yashmehrotra: > Without postgrest, would incident-commander even be functional?
Yeah, there cases for it. For example, we might want to run it in a cluster in agent-mode just to get data on pods and nodes, in that scenario postgrest does not serve any function.

Postgrest is only required when we run the UI

adityathebe: clarified with Yash. Will make that change.

adityathebe: > `http://localhost:<port>` (to start postgrest on the user given port)

@yashmehrotra A slight problem is that I run my own postgrest instance locally and then pass that as an externalURL to incident commander. The server is running locally but I don't want incident commander to fire its own postgrest. wdyt?

I don't remember exactly why I am running standalone postgrest externally, but that has been my local setup since the beginning.

yashmehrotra: umm, use `http://0.0.0.0:port` to pass that?

adityathebe: That'd work for me but seems hacky to design it like that.

yashmehrotra: Hows that hacky ?

adityathebe: I think, it's natural to supply "localhost"

yashmehrotra: The 0.0.0.0 suggestion is for your local to be working with this

PR #423, opened 2023-06-29
feat: add job history for upstream sync

yashmehrotra:


upstream/reconcile.go
```diff
@@ -8,27 +8,36 @@ import (
...
+	"github.com/flanksource/duty/models"
 	"github.com/flanksource/incident-commander/api"
 	"github.com/flanksource/incident-commander/db"
 )
 
 // SyncWithUpstream sends all the missing resources to the upstream.
 func SyncWithUpstream(ctx context.Context) error {
+	jobHistory := models.NewJobHistory("SyncWithUpstream", "", "")
+	_ = db.PersistJobHistory(jobHistory.Start())
+	defer db.PersistJobHistory(jobHistory.End())
```
adityathebe, line 20: This needs to be wrapped in an anonymous function.

yashmehrotra: Ahh, good catch

adityathebe (approved):

upstream/reconcile.go
```diff
@@ -8,27 +8,38 @@ import (
...
 	"net/url"
 
+	"github.com/flanksource/duty/models"
 	"github.com/flanksource/incident-commander/api"
 	"github.com/flanksource/incident-commander/db"
 )
 
 // SyncWithUpstream sends all the missing resources to the upstream.
 func SyncWithUpstream(ctx context.Context) error {
+	jobHistory := models.NewJobHistory("SyncWithUpstream", "", "")
```
moshloop, line 18: Add the upstream name

PR #436, opened 2023-07-07
feat: new notification system

adityathebe:


events/notifications.go
```diff
@@ -2,71 +2,129 @@ package events
...
 	"github.com/flanksource/incident-commander/notification"
 	"github.com/google/uuid"
-	"gorm.io/gorm"
 )
 
-func addNotificationEvent(ctx *api.Context, tx *gorm.DB, responder api.Responder) error {
-	var notifications []api.Notification
-	if err := tx.WithContext(ctx).Where("team_id = ? AND deleted_at IS NULL", responder.TeamID).Find(&notifications).Error; err != nil {
+func addNotificationEvent(ctx *api.Context, event api.Event) error {
+	// NOTE: might need to cache all the notifications instead of making this request on every event.
```
moshloop, line 14: We would definitely need to cache and compile the filters for the notifications

events/notifications.go
```diff
@@ -1,71 +1,289 @@
...
+					Name: EventNotificationPublish,
+					Properties: map[string]string{
+						"internal.message": data.Message,
+						"internal.url": fmt.Sprintf("smtp://%s:%s@%s:%s/?auth=Plain&FromAddress=%s&ToAddresses=%s",
+							url.QueryEscape(os.Getenv("SMTP_USER")),
+							url.QueryEscape(os.Getenv("SMTP_PASSWORD")),
+							os.Getenv("SMTP_HOST"),
+							os.Getenv("SMTP_PORT"),
+							url.QueryEscape(os.Getenv("SMTP_USER")),
+							url.QueryEscape(email),
```
adityathebe, line 100: @moshloop, do you think it's alright to have credentials in the event queue? If not then we can encrypt it.

Notification config in team spec or custom notifications can also have credentials in the URL (if they do not use connection name).

events/notifications.go
```diff
@@ -1,71 +1,340 @@
...
+		return fmt.Errorf("error templating notification: %w", err)
+	}
+
+	if props.PersonID != "" {
+		var emailAddress string
+		if err := ctx.DB().Model(&models.Person{}).Select("email").Where("id = ?", props.PersonID).Find(&emailAddress).Error; err != nil {
+			return fmt.Errorf("failed to get email of person(id=%s); %v", props.PersonID, err)
+		}
+
+		url := fmt.Sprintf("smtp://%s:%s@%s:%s/?auth=Plain&FromAddress=%s&ToAddresses=%s",
```
moshloop, line 86: We can't add the smtp details into the event message - I would just have a `email` event

notification/notification.go
```diff
@@ -1,8 +1,72 @@
...
+func GetNotifications(ctx *api.Context, eventName string) ([]models.Notification, error) {
+	if val, found := notificationByEventCache.Get(eventName); found {
+		return val.([]models.Notification), nil
+	}
+
+	var notifications []models.Notification
+	if err := ctx.DB().Where("deleted_at IS NULL").Where("? = ANY(events)", eventName).Find(&notifications).Error; err != nil {
+		return nil, err
+	}
+	notificationByEventCache.Set(eventName, notifications, cache.DefaultExpiration)
```
moshloop, line 33: We should compile the filter expressions and put them in the cache, so that they are not recompiled for every event.

events/notifications.go
adityathebe: @moshloop We are not adding it to the event. We're publishing it directly.

notification/notification.go
adityathebe: all expressions are cached by default in `utils/expression.go`

adityathebe:
- failure to parse the filter expression must be a fatal error

adityathebe:
image

PR #458, opened 2023-07-24
feat: multi-consumer event handling

yashmehrotra:


events/consumers.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+	}
+}
+
+func StartConsumers(gormDB *gorm.DB, config Config) {
+	allConsumers := []EventConsumer{
+		NewTeamConsumer(gormDB),
+		NewNotificationConsumer(gormDB),
+		NewResponderConsumer(gormDB),
+		NewUpstreamPushConsumer(gormDB, config),
```
adityathebe, line 191: I think we should not be setting up consumer for upstream if the upstream config is not fully provided.
Right now it panics [here](https://github.com/flanksource/incident-commander/pull/458/files#diff-5f3bc2bbe3b2795d2d392d7d394a13f7c4d422de3ae1a45fc7371d363cfa6402R29)

yashmehrotra: That was intentional, it will panic if we do not setup the agent mode but still receive push events.

Should I change it to not creating one ?

yashmehrotra: Have updated 👍🏽

yashmehrotra:
> @yashmehrotra incident-commander terminates immediately for me. I don't see any errors or panic in the logs https://pastebin.com/raw/H14GHsjF

Try again with the latest commit once

adityathebe:
@yashmehrotra still terminates immediately. That log I shared was when I had commented out upstream consumer.

yashmehrotra:
@adityathebe Does it work for you in the latest main branch ? Didn't change much about startup

Is it also crashing if you comment out the StartConsumers func in cmd/server ?

adityathebe:
@yashmehrotra main branch was failing for me as well. kubec config was incorrectly set for me and I think that caused the issue. Now it works fine.

events/consumers.go
```diff
@@ -0,0 +1,199 @@
...
+		if err != nil {
+			return fmt.Errorf("error acquiring database connection: %v", err)
+		}
+		defer conn.Release()
+
+		_, err = conn.Exec(ctx, "LISTEN event_queue_updates")
+		if err != nil {
+			return fmt.Errorf("error listening to database notifications: %v", err)
+		}
+		logger.Infof("listening to database notifications")
```
moshloop, line 133:
```suggestion
		logger.Debugf("listening to database notifications")
```

events/consumers.go
```diff
@@ -0,0 +1,199 @@
...
+
+		for {
+			notification, err := conn.Conn().WaitForNotification(ctx)
+			if err != nil {
+				return fmt.Errorf("error listening to database notifications: %v", err)
+			}
+
+			logger.Tracef("Received database notification: %+v", notification)
```
moshloop, line 141:
```suggestion
```

events/notifications.go
```diff
@@ -13,8 +13,56 @@ import (
 	pkgNotification "github.com/flanksource/incident-commander/notification"
 	"github.com/flanksource/incident-commander/teams"
 	"github.com/google/uuid"
+	"gorm.io/gorm"
 )
 
+func NewNotificationConsumer(db *gorm.DB) EventConsumer {
+	return EventConsumer{
+		WatchEvents: []string{
+			EventNotificationSend, EventNotificationUpdate, EventNotificationDelete,
```
moshloop, line 22: I think event notification send should be its own consumer with 5 consumers - these can potentially be slow

moshloop: I think the event is duplicated now ?

moshloop:
@yashmehrotra can you check the failing tests ?

events/notifications.go
yashmehrotra: Added validation for that as well

yashmehrotra:
> @yashmehrotra can you check the failing tests ?

@moshloop Fixed and rebased, all is green now

PR #471, opened 2023-08-02
feat: support for clerk auth 

yashmehrotra:
Merge https://github.com/flanksource/mission-control-chart/pull/98 after this PR is merged


auth/clerk_client.go
```diff
@@ -0,0 +1,161 @@
...
+
+		// Extract session token from Authorization header
+		sessionToken := c.Request().Header.Get(echo.HeaderAuthorization)
+		sessionToken = strings.TrimPrefix(sessionToken, "Bearer ")
+
+		var (
+			user   *clerk.User
+			err    error
+			sessID string
+		)
```
adityathebe, line 53: nit: this can go away

auth/clerk_client.go
```diff
@@ -0,0 +1,161 @@
...
+				logger.Errorf("Error creating user in database from clerk: %v", err)
+				return c.String(http.StatusUnauthorized, "Unauthorized")
+			}
+			// Clear user from cache so that new metadata is used
+			h.userCache.Delete(sessID)
+		}
+
+		token, err := h.getDBToken(sessID, *user.ExternalID)
+		if err != nil {
+			logger.Errorf("Error generating JWT Token: %v", err)
```
adityathebe, line 73: probably needs to return as well

auth/clerk_client.go
```diff
@@ -0,0 +1,161 @@
...
+
+func (h *ClerkHandler) getUser(sessionToken string) (*clerk.User, string, error) {
+	sessClaims, err := h.client.VerifyToken(sessionToken)
+	if err != nil {
+		return nil, "", err
+	}
+
+	cacheKey := sessClaims.SessionID
+	if user, exists := h.userCache.Get(cacheKey); exists {
+		return user.(*clerk.User), "", nil
```
adityathebe, line 112: don't we need to return session id in here?

auth/clerk_client.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func (h ClerkHandler) Session(next echo.HandlerFunc) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		if canSkipAuth(c) {
+			return next(c)
+		}
+
+		// Extract session token from Authorization header
+		sessionToken := c.Request().Header.Get(echo.HeaderAuthorization)
```
moshloop, line 46: I think we might also need to extract it via the `_session` cookie

auth/clerk_client.go
yashmehrotra: Good catch

auth/clerk_client.go
yashmehrotra: 👍🏽

PR #475, opened 2023-08-03
feat: create agent api

adityathebe:
Resolves: https://github.com/flanksource/incident-commander/issues/465


auth/middleware.go
```diff
@@ -75,6 +81,33 @@ func (k *kratosMiddleware) Session(next echo.HandlerFunc) echo.HandlerFunc {
 	}
 }
 
+func (k *kratosMiddleware) validateAccessToken(ctx context.Context, username, password string) (*models.AccessToken, error) {
```
adityathebe, line 84: This will be adjusted once Yash's PR on clerk integration is merged

db/people.go
```diff
@@ -22,3 +26,24 @@ func UpdateUserProperties(ctx *api.Context, userID string, newProps api.PersonPr
...
+
+	person := models.Person{Name: username, Type: "agent"}
+	if err := tx.Clauses(clause.Returning{}).Create(&person).Error; err != nil {
+		return nil, err
+	}
+
+	accessToken := models.AccessToken{
+		Value:     hashedPassword,
+		PersonID:  person.ID,
+		ExpiresAt: time.Now().Add(time.Hour), // TODO: decide on this one
```
adityathebe, line 42: If we can have a never expiring token then will leave this as nil.

db/people.go
```diff
@@ -22,3 +26,25 @@ func UpdateUserProperties(ctx *api.Context, userID string, newProps api.PersonPr
...
+	person := models.Person{Name: username, Type: "agent"}
+	if err := tx.Clauses(clause.Returning{}).Create(&person).Error; err != nil {
+		return nil, err
+	}
+
+	accessToken := models.AccessToken{
+		Name:      "default",
+		Value:     hashedPassword,
+		PersonID:  person.ID,
+		ExpiresAt: time.Now().Add(time.Hour * 24 * 90), // long-lived token
```
moshloop, line 43: This is too short - Make it 365 days

db/people.go
```diff
@@ -22,3 +26,25 @@ func UpdateUserProperties(ctx *api.Context, userID string, newProps api.PersonPr
...
+func CreatePerson(ctx *api.Context, username, hashedPassword string) (*models.Person, error) {
+	tx := ctx.DB().Begin()
+	defer tx.Rollback()
+
+	person := models.Person{Name: username, Type: "agent"}
+	if err := tx.Clauses(clause.Returning{}).Create(&person).Error; err != nil {
+		return nil, err
+	}
+
+	accessToken := models.AccessToken{
```
moshloop, line 39: I would wrap this in a db.CreateAccessToken(person, name, expiry) that generates, then saves the token returning the plain text token

db/people.go
```diff
@@ -22,3 +26,25 @@ func UpdateUserProperties(ctx *api.Context, userID string, newProps api.PersonPr
...
+		Name:      "default",
```
moshloop, line 40: This should be "agent-" timestamp

rbac/init.go
```diff
@@ -124,6 +127,8 @@ func Init(adminUserID string) error {
 		{RoleResponder, ObjectDatabaseIncident, ActionUpdate},
 
 		{RoleViewer, ObjectDatabase, ActionRead},
+
+		{RoleAgent, ObjectAgentPush, ActionWrite},
```
moshloop, line 131: The agent should only have access to call the API that would be used when pushing upstream

db/agents.go
```diff
@@ -50,3 +51,13 @@ func GetOrCreateAgent(ctx *api.Context, name string) (*models.Agent, error) {
 
 	return a, nil
 }
+
+func CreateAgent(ctx *api.Context, name string, personID *uuid.UUID, properties map[string]string) error {
+	a := models.Agent{
+		Name:       name,
+		PersonID:   personID,
+		Properties: properties,
```
moshloop, line 59: Need a type=agent

auth/middleware.go
```diff
@@ -83,6 +116,25 @@ func (k *kratosMiddleware) validateSession(r *http.Request) (*client.Session, er
 	}
 
 	if username, password, ok := r.BasicAuth(); ok {
+		if strings.HasPrefix(username, "agent-") {
```
moshloop, line 119:
```suggestion
		if username == "TOKEN" {
```

auth/middleware.go
```diff
@@ -83,6 +116,25 @@ func (k *kratosMiddleware) validateSession(r *http.Request) (*client.Session, er
...
+			accessToken, err := k.validateAccessToken(r.Context(), username, password)
```
moshloop, line 120: username is meaningless

auth/middleware.go
```diff
@@ -75,6 +81,33 @@ func (k *kratosMiddleware) Session(next echo.HandlerFunc) echo.HandlerFunc {
...
+  access_tokens.*
+FROM
+  access_tokens
+  LEFT JOIN people ON access_tokens.person_id = people.id
+WHERE
+  people.name = ?
+  AND access_tokens.expires_at > NOW()`
+
+	var acessToken models.AccessToken
+	if err := k.db.Raw(query, username).First(&acessToken).Error; err != nil {
```
moshloop, line 96: This is the other way around, we bcrypt the token and then do a lookup for the hashed value to get the person and expiry

moshloop: If the token has expired that should be included in the error message returned - "token expired on ABC"

agent/agent.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	if err := db.CreateAgent(ctx, body.Name, &person.ID, body.Properties); err != nil {
+		return nil, fmt.Errorf("failed to create a new agent: %w", err)
+	}
+
+	return &api.GeneratedAgent{
+		ID:          person.ID.String(),
+		Username:    username,
+		AccessToken: password,
```
moshloop, line 41: The token will need to include the generated salt/cost, so that the hashed token can be generated from it  and then looked up in the db

db/agents.go
adityathebe: @moshloop why have type=agent in an agents table? What would other types be?

auth/middleware.go
adityathebe: > we bcrypt the token and then do a lookup for the hashed value
Current std library doesn't allow creating a bcrypt hash with a custom salt. We'd need to resort to a different hashing algorithm.

Although I think what we currently have is fine (apart from the expiry clause in the SQL query).

adityathebe: We'll go with argon2

rbac/init.go
adityathebe: Isn't it already that way?

agent/agent.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return "", "", err
+	}
+
+	password, err = rand.GenerateRandHex(32)
+	if err != nil {
+		return "", "", err
+	}
+
+	return fmt.Sprintf("agent-%s", username), password, nil
```
yashmehrotra, line 57: use constant here

adityathebe: usernames are irrelevant and the prefix is never reused.

utils/utils.go
```diff
@@ -0,0 +1,5 @@
+package utils
```
yashmehrotra, line 1: Please move this to commons, I know a small patch + release is a hassle 😅  but we have this func in all repos

db/people.go
```diff
@@ -34,3 +42,52 @@ func GetOrCreateUser(ctx *api.Context, user api.Person) (api.Person, error) {
...
+		Value:     encodedHash,
+		PersonID:  personID,
+		ExpiresAt: time.Now().Add(expiry), // long-lived token
+	}
+	if err := ctx.DB().Create(&accessToken).Error; err != nil {
+		return "", err
+	}
+
+	formattedHash := fmt.Sprintf("%s.%s.%d.%d.%d", password, salt, timeCost, memoryCost, parallelism)
+	return formattedHash, nil
```
yashmehrotra, line 92: Should we store hash generation meta in the database ? If we change the algo/logic later on, it would be easier to support older logic

adityathebe: If we want to prepare ourselves for any change in algo in future, then I think we can add a suffix that can identify the algo used to generate the token.

@moshloop what do you think?

cmd/server.go
```diff
@@ -141,12 +143,14 @@ func createHTTPServer(gormDB *gorm.DB) *echo.Echo {
...
 		logger.Warnf("Please ensure that all the required flags for upstream is supplied.")
 	}
-	upstreamGroup := e.Group("/upstream")
+	upstreamGroup := e.Group("/upstream", rbac.Authorization(rbac.ObjectAgentPush, rbac.ActionWrite))
 	upstreamGroup.POST("/push", upstream.PushUpstream)
 	upstreamGroup.GET("/pull/:agent_name", upstream.Pull)
 	upstreamGroup.GET("/canary/pull/:agent_name", canary.Pull)
 	upstreamGroup.GET("/status/:agent_name", upstream.Status)
 
+	e.POST("/agent/generate", agent.GenerateAgent)
```
adityathebe, line 152: @moshloop Should this be added to exclusion in the kratos middleware?
In here https://github.com/flanksource/incident-commander/blob/main/auth/middleware.go#L48

moshloop: No - generating an agent should be restricted to the admin role

PR #483, opened 2023-08-05
refactor: process upstream messages individually if push fails

yashmehrotra:


events/upstream_push.go
```diff
@@ -150,12 +150,25 @@ func (t *pushToUpstreamEventHandler) Run(ctx *api.Context, events []api.Event) [
 	}
 
 	upstreamMsg.ApplyLabels(t.conf.LabelsMap())
-	if err := upstream.Push(ctx, t.conf, upstreamMsg); err != nil {
+	err := upstream.Push(ctx, t.conf, upstreamMsg)
+	if err == nil {
+		return failedEvents
+	}
+
+	if len(events) == 1 {
```
adityathebe, line 158: I am not following why a single event is a failure

adityathebe: oh nevermind, the error comes from `upstream.Push`

events/upstream_push.go
```diff
@@ -150,12 +150,25 @@ func (t *pushToUpstreamEventHandler) Run(ctx *api.Context, events []api.Event) [
...
 		errMsg := fmt.Errorf("failed to push to upstream: %w", err)
 		failedEvents = append(failedEvents, addErrorToFailedEvents(events, errMsg)...)
+		return failedEvents
+	} else {
+		// Error encountered while pushing could be an SQL or Application error
+		// Since we do not know which event in the bulk is failing
+		// Process each event individually since upsteam.Push is idempotent
+		var failedIndividualEvents []api.Event
+		for _, e := range events {
+			failedIndividualEvents = append(failedIndividualEvents, t.Run(ctx, []api.Event{e})...)
```
adityathebe, line 168: I think for the sake of simplicity we could avoid the recursive call in here. It would probably work in similar way I assume.

adityathebe: I now see that `Run` would otherwise receive a group of events, so `t.Run` makes sense here. 
Maybe we need to do the same for `failedEvents`. 

Or in general, any failed event should be handled individually. What do you think of modifying `consumeEvents()` so when it's fetching in batches, it's making sure it's not pulling any failed events. And if it is pulling any failed event it's doing that one at a time.

yashmehrotra: In all other consumers, although we process events in bulk, each event passed or fails individually in isolation. There is no need to handle failed events separately there

For upstream push, we are mixing N events into one function call, the entirety of which can fail leading to us not knowing which event was the culprit, this is why in that scenario, we are just reprocessing all again individually so that the other successful events don't get marked as failed due to the failed events

In other consumers, this is already handled in the implementation since events do not get mixed

PR #498, opened 2023-08-14
feat: add an event_queue cleaning job

adityathebe:
Resolves: https://github.com/flanksource/incident-commander/issues/411

### DIFF
--- a/jobs/event_queue.go
+++ b/jobs/event_queue.go
@@ -1,0 +1,48 @@
+package jobs
+
+import (
+	"time"
+
+	"github.com/flanksource/commons/logger"
+	"github.com/flanksource/duty/models"
+	"github.com/flanksource/incident-commander/api"
+	"github.com/flanksource/incident-commander/db"
+)
+
+// CleanupEventQueue deletes state records in the `event_queue` table
+func CleanupEventQueue() {
+	ctx := api.NewContext(db.Gorm, nil)
+
+	jobHistory := models.NewJobHistory("CleanupEventQueue", "", "")
+	_ = db.PersistJobHistory(ctx, jobHistory.Start())
+
+	pushQueueSchedule := map[string]time.Duration{
+		"topologies":                     time.Hour * 24 * 7,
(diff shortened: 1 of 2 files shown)


jobs/event_queue.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"time"
+
+	"github.com/flanksource/commons/logger"
+	"github.com/flanksource/duty/models"
+	"github.com/flanksource/incident-commander/api"
+	"github.com/flanksource/incident-commander/db"
+)
+
+// CleanupEventQueue deletes state records in the `event_queue` table
```
yashmehrotra, line 12: stale*

jobs/event_queue.go
```diff
@@ -0,0 +1,48 @@
...
+func CleanupEventQueue() {
```
yashmehrotra, line 13: Rename to CleanupPushEvents or something

yashmehrotra (approved):
LGTM

jobs/event_queue.go
adityathebe: I think we would want to clean up other event types in the queue as well.

adityathebe: Added deletion of other events as well

jobs/event_queue.go
```diff
@@ -0,0 +1,58 @@
...
+		"check_statuses":                 time.Hour * 24 * 7,
+	}
+
+	for table, age := range pushQueueSchedule {
+		result := db.Gorm.Debug().Exec("DELETE FROM event_queue WHERE name = 'push_queue.create' AND properties->>'table' = ? AND NOW() - created_at > ?", table, age)
+		if result.Error != nil {
+			logger.Errorf("Error cleaning up push_queue events for table=%s: %v", table, result.Error)
+			jobHistory.AddError(result.Error.Error())
+		} else {
+			logger.Debugf("Cleaned up %d push_queue events for table=%s", result.RowsAffected, table)
```
moshloop, line 40: This should probably be a Warn if row coount > 0

PR #499, opened 2023-08-15
feat: playbooks

adityathebe:


playbook/controllers.go
```diff
@@ -0,0 +1,89 @@
...
+
+	if r.ConfigID != uuid.Nil && r.ComponentID != uuid.Nil {
+		return fmt.Errorf("either config_id or component_id is required")
+	}
+
+	return nil
+}
+
+// HandlePlaybookRun handles playbook run requests.
+func HandlePlaybookRun(c echo.Context) error {
```
moshloop, line 43: I think this should queue the playbook run, with a startdate of now and then we have a background process that runs them

playbook/controllers.go
```diff
@@ -0,0 +1,134 @@
...
+	var req RunParams
+	if err := c.Bind(&req); err != nil {
+		return c.JSON(http.StatusBadRequest, api.HTTPError{Error: err.Error(), Message: "invalid request"})
+	}
+
+	if err := req.valid(); err != nil {
+		return c.JSON(http.StatusBadRequest, api.HTTPError{Error: err.Error(), Message: "invalid request"})
+	}
+
+	playbook, err := db.FindPlaybook(ctx, req.ID)
```
moshloop, line 81: I think the execution of the playbook should be in a separate file/function - especially so that it can be tested.

adityathebe: The function that executes is isolated. This controller simply records the run in the db.

adityathebe:
@moshloop This is ready for review now.
- implemented approvals by a person/team
- endpoint that lists all the suitable playbooks for a config/component

adityathebe:
@moshloop @yashmehrotra I've added a new application level error struct that can be easily mapped to HTTP errors. It was kind of necessary for me right now so I've added it in this PR itself. 

It's pretty much a copy of https://github.com/benbjohnson/wtf/blob/main/error.go with some minor changes.

playbook/playbook.go
```diff
@@ -0,0 +1,30 @@
...
+
+func ListPlaybooksOfComponent(ctx *api.Context, id string) ([]models.Playbook, error) {
+	var component models.Component
+	if err := ctx.DB().Where("id = ?", id).Find(&component).Error; err != nil {
+		return nil, err
+	} else if component.ID == uuid.Nil {
+		return nil, api.Errorf(api.ENOTFOUND, "component(id=%s) not found", id)
+	}
+
+	return db.FindPlaybooksByTypeAndTags(ctx, component.Type, component.Labels)
```
moshloop, line 29: I think this will. need to be another function and look at components and not configs

playbook/playbook_test.go
```diff
@@ -0,0 +1,201 @@
...
+			}
+
+			attempts += 1
+			if attempts > 20 { // wait for 2 seconds
+				ginkgo.Fail(fmt.Sprintf("Timed out waiting for run to be scheduled. Status = %s", savedRun.Status))
+			}
+		}
+	})
+
+	ginkgo.It("should execute playbook", func() {
```
moshloop, line 169: Can we add tests for playbook selection on both components and configs

playbook/playbook_test.go
```diff
@@ -0,0 +1,201 @@
...
+			Spec:   spec,
+			Source: models.SourceConfigFile,
+		}
+
+		err = testDB.Create(&playbook).Error
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	ginkgo.It("should store playbook run via API", func() {
+		run := RunParams{
```
moshloop, line 85: I don't think it is necessary to call these functions via HTTP - we can all the functions directly

playbook/queue_consumer.go
```diff
@@ -0,0 +1,159 @@
...
+		tickInterval:                   time.Minute,
+		dbReconnectMaxDuration:         time.Minute,
+		dbReconnectBackoffBaseDuration: time.Second,
+		registry:                       sync.Map{},
+	}
+}
+
+func (t *queueConsumer) Listen() error {
+	pgNotify := make(chan string)
+	go utils.ListenToPostgresNotify(t.pool, "playbook_run_updates", t.dbReconnectMaxDuration, t.dbReconnectBackoffBaseDuration, pgNotify)
```
moshloop, line 45: I would move this method to the QueueConsumer .e.g

t.PgListen("playbook_spec_approval", approvalUpdated)

playbook/queue_consumer.go
```diff
@@ -0,0 +1,159 @@
...
+
+		case <-time.After(t.tickInterval):
+			if err := t.consumeAll(ctx); err != nil {
+				logger.Errorf("%v", err)
+			}
+		}
+	}
+}
+
+func (t *queueConsumer) onPlaybookSpecApprovalUpdated(ctx *api.Context, playbookID string) error {
```
moshloop, line 79: These functions are all in the playbook package so no need to duplicate prefix
```suggestion
func (t *queueConsumer) onApprovalUpdated(ctx *api.Context, playbookID string) error {
```

playbook/queue_consumer.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	if len(runs) == 0 {
+		return nil
+	}
+
+	for _, r := range runs {
+		go func(run models.PlaybookRun) {
+			if _, loaded := t.registry.LoadOrStore(run.ID, nil); !loaded {
+				if !run.StartTime.After(time.Now()) {
```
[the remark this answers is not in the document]
moshloop: I don't think we should be sleeping inside go functions - we also shouldn't be comparing server time and db time. 

What is wrong with `select * from runs where starttime >= now() and status = 'pending'` ?

Can we not re-use the queue consumer for events ? as it is a very similar process, just on a different table

api/errors.go
```diff
@@ -0,0 +1,86 @@
...
+
+// WithDebugInfo wraps an application error with a debug message.
+func (e *Error) WithDebugInfo(msg string, args ...any) *Error {
+	e.DebugInfo = fmt.Sprintf(msg, args...)
+	return e
+}
+
+// ErrorCode unwraps an application error and returns its code.
+// Non-application errors always return EINTERNAL.
+func ErrorCode(err error) string {
```
moshloop, line 47: This should probably be a WithError

playbook/queue_consumer.go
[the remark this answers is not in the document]
adityathebe: > select * from runs where starttime >= now() and status = 'pending'

In this case we could be fetching runs which are schedule in future. And it wouldn't execute runs whose start time was missed.

playbook/runner.go
```diff
@@ -0,0 +1,137 @@
...
+			return fmt.Errorf("action %s failed: %w", action.Name, err)
+		}
+	}
+
+	return nil
+}
+
+func executeAction(ctx *api.Context, run models.PlaybookRun, action v1.PlaybookAction, actionParam ActionParam) ([]byte, error) {
+	if action.Exec != nil {
+		script, err := gomplate.RunTemplate(actionParam.AsMap(), gomplate.Template{Template: action.Exec.Script})
```
moshloop, line 121: I think this should be in the exec.go file

playbook/runner.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+		action.Exec.Script = script
+
+		e := actions.ExecAction{}
+		res, err := e.Run(ctx, *action.Exec)
+		if err != nil {
+			return nil, err
+		}
+
+		return json.Marshal(res.Stdout)
```
moshloop, line 133: I think we need returncode, stdout and stderr in a json object - and they need to be saved backed to the playbook_run_action.result column

fixtures/playbooks/deleting-configmap.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: mission-control.flanksource.com/v1
+kind: Playbook
+metadata:
+  name: delete-kubernetes-configmap
+spec:
+  description: Delete Kubernetes ConfigMap
+  configs:
+    - type: Kubernetes::ConfigMap
+      tags:
+        namespace: default
```
moshloop, line 10: namespace should come from the config itself

api/errors.go
adityathebe: `WithError` would make sense if we were attaching the error. But here we're just extracting the code from the error.

playbook/queue_consumer.go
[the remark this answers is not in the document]
adityathebe: The scheduled time doesn't have to be accurate. It's fine to have an offset of a minute.

playbook/playbook_test.go
adityathebe: Everything about saving a playbook run is handled by the HTTP handler. The actual task is simply to save to DB and the handler performs other validation on top of that.

I think its better to test the handler this way than just directly saving to db from the test.

PR #520, opened 2023-08-28
refactor: event consumer

adityathebe:


playbook/run_consumer.go
```diff
@@ -0,0 +1,50 @@
...
+func NewEventQueueConsumer(db *gorm.DB, pool *pgxpool.Pool) *events.EventConsumer {
+	return events.NewEventConsumer(db, pool, "playbook_run_updates", EventConsumer).WithNumConsumers(5)
+}
+
+func EventConsumer(ctx *api.Context, batchSize int) error {
+	tx := ctx.DB().Begin()
+	if tx.Error != nil {
+		return fmt.Errorf("error initiating db tx: %w", tx.Error)
+	}
+	defer tx.Rollback()
```
yashmehrotra, line 22: Is this safe to run after doing a `tx.Commit()` ? does it fail silently in that case ?

events/event_queue.go
```diff
@@ -0,0 +1,174 @@
...
+
+	for i := range allConsumers {
+		go allConsumers[i].Listen()
+	}
+}
+
+// newEventQueueConsumerFunc returns a new event consumer for the `event_queue` table
+// based on the given watch events and process batch function.
+func newEventQueueConsumerFunc(watchEvents []string, processBatchFunc ProcessBatchFunc) EventConsumerFunc {
+	return func(ctx *api.Context, batchSize int) error {
```
yashmehrotra, line 112: Whats the advantage of returning this as a function vs calling the function directly ?

cmd/server.go
```diff
@@ -231,13 +231,15 @@ var Serve = &cobra.Command{
 
 		go jobs.Start()
 
-		events.StartConsumers(db.Gorm, events.Config{
+		events.StartConsumers(db.Gorm, db.Pool, events.Config{
 			UpstreamPush: api.UpstreamConf,
 		})
 
-		playbookRunConsumer := playbook.NewQueueConsumer(db.Gorm, db.Pool)
+		playbookRunConsumer := playbook.NewEventQueueConsumer(db.Gorm, db.Pool)
```
yashmehrotra, line 238: Should we name his RunsConsumer or something with `run` in the name  ?

events/event_queue.go
yashmehrotra: Is it for passing a different consumer func to playbook runs?

playbook/run_consumer.go
adityathebe: yea. Once it's committed, the rollback is a no-op.

events/event_queue.go
adityathebe: Ya, the `EventConsumer` needed to be generic and `watchEvents` and `processBatchFunc` are specific to consumers of `event_queue` and doesn't fit well with playbook even consumer.

cmd/server.go
adityathebe: I see. And call the `.Listen` inside it?

yashmehrotra: yeah, something like that

adityathebe:
@moshloop Can we merge this?

playbook/pg_listeners.go
```diff
@@ -0,0 +1,89 @@
...
+		case id := <-pgNotifyPlaybookApprovalsInserted:
+			if err := onPlaybookRunNewApproval(ctx, id); err != nil {
+				logger.Errorf("%v", err)
+			}
+		}
+	}
+}
+
+// onApprovalUpdated is called when the playbook spec approval is updated
+func onApprovalUpdated(ctx *api.Context, playbookID string) error {
```
moshloop, line 46: Is this the only place the approvals are managed? We can't use pg_notify as the only source as we will miss event when incident commander is down

adityathebe: Right! I'll add new sync consumers but sync consumers are implemented in a later PR. I'll make the change there.

notification/notification_test.go
```diff
@@ -104,13 +104,15 @@ var _ = ginkgo.Describe("Test Notification on incident creation", ginkgo.Ordered
 	})
 
 	ginkgo.It("should consume the event and send the notification", func() {
-		notifHandler := events.NewNotificationConsumer(db.Gorm)
-		sendHandler := events.NewNotificationSendConsumer(db.Gorm)
+		notifHandler := events.NewNotificationConsumer(db.Gorm, db.Pool)
```
moshloop, line 107:
```suggestion
		notificationHandler := events.NewNotificationConsumer(db.Gorm, db.Pool)
```

PR #521, opened 2023-08-28
feat: sync consumers

adityathebe:


events/event_queue.go
```diff
@@ -54,28 +59,18 @@ type Config struct {
 // consumerWatchEvents keeps a registry of all the event_queue consumer and the events they watch.
 // This helps in ensuring that a single event is not being consumed by multiple consumers.
 var consumerWatchEvents = map[string][]string{
```
moshloop, line 61: Can we create a  new type to make the behaviour explicit e.g

SyncHandler('queue)
AsyncHandler('a','b')

events/event_queue.go
```diff
@@ -169,6 +162,26 @@ func newEventQueueConsumerFunc(watchEvents []string, processBatchFunc ProcessBat
 			}
 		}
 
+		// Process sync consumers
+		if len(syncConsumers) != 0 {
+			failedEventsIDs := set.New[uuid.UUID]()
```
moshloop, line 167: This doesn't make sense for sync consumers? - I think its better to have 2 different functions one for sync and 1 for async

events/event_queue.go
adityathebe: We do have a new type https://github.com/flanksource/incident-commander/pull/521/files/e16dd997b12c9ac6c38faa9cfb0c1f1add86064a#diff-2010347d26487a6fd1e74fc8c511d544b89eeef547356ad7c40d7bd7e88fa285R23

adityathebe:
Merge after: 
- https://github.com/flanksource/incident-commander/pull/520
- https://github.com/flanksource/duty/pull/260

auth/middleware.go
```diff
@@ -102,7 +102,10 @@ func (k *kratosMiddleware) Session(next echo.HandlerFunc) echo.HandlerFunc {
 				email = e
 			}
 		}
-		ctx.WithUser(&api.ContextUser{ID: uuid.MustParse(session.Identity.GetId()), Email: email})
+
+		if uid, err := uuid.Parse(session.Identity.GetId()); nil == err {
```
yashmehrotra, line 106: nit: `err == nil`

yashmehrotra: If parsing fails, 4xx error unauthorized should be returned, auth being passed means that a user-id is expected in further functions

adityathebe: > `nit: err == nil`

This is actually deliberate. `err != nil` is so prevalent in Go that even `err == nil` might appear as `err != nil` on a glance. `nil == err` makes it noticeable in my opinion.

> If parsing fails, 4xx error unauthorized should be returned, auth being passed means that a user-id is expected in further functions

Had to skip auth on `/kratos/*` calls other wise if we return 401, the login page would break. I am not entirely sure if this is okay.

image

yashmehrotra: Can you confirm with @mainawycliffe once how frontend should behave in these scenarios

mainawycliffe: @yashmehrotra @adityathebe There are a few endpoints I would expect to be public - login, registration (not currently in use), password reset, etc., but others such as user profile related would probably need to be authenticated.

adityathebe: @yashmehrotra you think it's fine to skip auth on all `/kratos/*` calls?

events/playbook.go
```diff
@@ -0,0 +1,11 @@
+package events
+
+import "github.com/flanksource/incident-commander/api"
+
+func SavePlaybookRun(ctx *api.Context, event api.Event) error {
```
moshloop, line 5:
```suggestion
func SchedulePlaybookRun(ctx *api.Context, event api.Event) error {
```

events/checks.go
```diff
@@ -0,0 +1,8 @@
+package events
+
+func NewCheckConsumerSync() SyncEventConsumer {
+	return SyncEventConsumer{
```
adityathebe, line 4: @moshloop I initially tried this in an `init()` func but the upstream consumer
- needs a custom config
- and also might not have to be initiated at all.

Due to that I had to stick with this constructor pattern. However, the uniqueness is enforced now.

auth/middleware.go
moshloop: > This is actually deliberate. err != nil is so prevalent in Go that even err == nil might appear as err != nil on a glance. nil == err makes it noticeable in my opinion.

I don't see why nil == err is any different from err == nil,  Go is meant to be un-opionated so that when you read code you don't need to think about formatting and nil == err, is non-standard and makes you think so even if it is technically better in isolation, its deviation from the standard makes it worse

PR #524, opened 2023-08-29
feat: Trigger playbooks on events

adityathebe:


events/playbook.go
```diff
@@ -1,11 +1,157 @@
...
+		}
+
+		if playbook.Spec.Approval == nil || playbook.Spec.Approval.Approvers.Empty() {
+			run.Status = models.PlaybookRunStatusScheduled
+		}
+
+		switch specEvent.Class {
+		case "canary":
+			// What should be the component or config id?
+			if ok, err := matchResource(eventResource.Check.Labels, map[string]any{"check": eventResource.Check}, playbook.Spec.On.Canary); err != nil {
```
moshloop, line 69: why not just take an eventResource object e.g. matchResource(eventResource, playbook.Spec.On.Canary)

fixtures/playbooks/scale-deployment.yaml
```diff
@@ -3,6 +3,18 @@ kind: Playbook
 metadata:
   name: scale-deployment
 spec:
+  'on':
```
moshloop, line 6: Add new fixtures for the event handling

events/playbook.go
adityathebe: `matchResource` does not know whether it should use the labels from the check or the component.

events/playbook.go
```diff
@@ -104,36 +107,38 @@ func logToJobHistory(ctx *api.Context, playbookID, err string) {
 // for the given labels and cel env.
 func matchResource(labels map[string]string, celEnv map[string]any, matchFilters []v1.PlaybookEventDetail) (bool, error) {
 	for _, mf := range matchFilters {
+		var (
```
moshloop, line 110: Can we not just use a `continue outer` with an `outer: ` label ?

PR #528, opened 2023-09-01
feat: set title from the notification if needed

adityathebe:
Resolves: https://github.com/flanksource/incident-commander/issues/513


adityathebe:
@moshloop This test will fail unless we merge the event consumer PR. This is using the latest duty which removes the pg notify for playbook approvals.

moshloop:
@adityathebe  please rebase

PR #529, opened 2023-09-01
feat: use default SMTP credential

adityathebe:
Resolves: https://github.com/flanksource/incident-commander/issues/514

### DIFF
--- a/notification/shoutrrr.go
+++ b/notification/shoutrrr.go
@@ -9,12 +9,14 @@ import (
 	"github.com/containrrr/shoutrrr"
 	"github.com/containrrr/shoutrrr/pkg/types"
 	"github.com/flanksource/commons/collections"
+	"github.com/flanksource/commons/utils"
 	"github.com/flanksource/incident-commander/api"
+	"github.com/flanksource/incident-commander/mail"
 )
 
 // defaultSMTPPrefix indicates that the shoutrrr URL for smtp should use
 // the system's SMTP credentials.
-const defaultSMTPPrefix = "smtp://auto:auto@auto/"
+const defaultSMTPPrefix = "smtp://system/"
 
 // setSystemSMTPCredential modifies the shoutrrrURL to use the system's SMTP credentials.
 func setSystemSMTPCredential(shoutrrrURL string) (string, error) {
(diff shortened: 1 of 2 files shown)


notification/shoutrrr.go
```diff
@@ -10,6 +12,33 @@ import (
...
+	)
+	shoutrrrURL = strings.ReplaceAll(shoutrrrURL, defaultSMTPPrefix, prefix)
+
+	parsedURL, err := url.Parse(shoutrrrURL)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse shoutrrr URL: %w", err)
+	}
+
+	query := parsedURL.Query()
+	query.Set("FromAddress", os.Getenv("SMTP_USER"))
```
moshloop, line 35: I think we are also missing an SMTP_FROM - note the user and from is not necessarily the same (if from not specified then use SMTP_USER)

notification/shoutrrr.go
```diff
@@ -10,6 +12,33 @@ import (
 	"github.com/flanksource/incident-commander/api"
 )
 
+// defaultSMTPPrefix indicates that the shoutrrr URL for smtp should use
+// the system's SMTP credentials.
+const defaultSMTPPrefix = "smtp://auto:auto@auto/"
```
moshloop, line 17:
```suggestion
const defaultSMTPPrefix = "smtp://system/"
```

notification/shoutrrr.go
adityathebe: We seem to be taking SMTP from as a flag. Will use that.

notification/shoutrrr.go
```diff
@@ -2,14 +2,45 @@ package notification
...
+	)
+	shoutrrrURL = strings.ReplaceAll(shoutrrrURL, defaultSMTPPrefix, prefix)
+
+	parsedURL, err := url.Parse(shoutrrrURL)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse shoutrrr URL: %w", err)
+	}
+
+	query := parsedURL.Query()
+	query.Set("FromAddress", utils.Coalesce(mail.FromAddress, os.Getenv("SMTP_FROM"), os.Getenv("SMTP_USER")))
```
moshloop, line 37: Will this not always return mail.FromAddress as it has a default?

notification/shoutrrr.go
```diff
@@ -2,14 +2,45 @@ package notification
...
 	"github.com/containrrr/shoutrrr/pkg/types"
 	"github.com/flanksource/commons/collections"
+	"github.com/flanksource/commons/utils"
 	"github.com/flanksource/incident-commander/api"
+	"github.com/flanksource/incident-commander/mail"
 )
 
+// defaultSMTPPrefix indicates that the shoutrrr URL for smtp should use
+// the system's SMTP credentials.
+const defaultSMTPPrefix = "smtp://system/"
```
moshloop, line 19: Can we create a SendEmail(ctx, title, message, properties) that uses this default? and then replace all other usages of SMTP_* with this method ?

notification/shoutrrr.go
adityathebe: on the rare cases where the flag is explicitly removed, this will handle it.

PR #535, opened 2023-09-06
feat: default title and body for notifications

adityathebe:
Resolves: https://github.com/flanksource/incident-commander/issues/533
Resvoles: https://github.com/flanksource/incident-commander/issues/532


adityathebe:
Will implement https://github.com/flanksource/incident-commander/issues/532 in this PR

adityathebe:
@moshloop I tried formatting the URLs but Shoutrrr doesn't support that at the moment (example: Slack).
And we would also need to specify whether to use markdown or HTML per service.

So, for now I've simply added raw URL in the default templates.

cmd/root.go
```diff
@@ -49,6 +49,7 @@ func ServerFlags(flags *pflag.FlagSet) {
 	flags.IntVar(&metricsPort, "metricsPort", 8081, "Port to expose a health dashboard ")
 	flags.BoolVar(&dev, "dev", false, "Run in development mode")
 	flags.StringVar(&publicEndpoint, "public-endpoint", "http://localhost:8080", "Public endpoint that this instance is exposed under")
+	flags.StringVar(&api.PublicWebURL, "public-web-url", "http://localhost:3000", "Public URL of the web application")
```
moshloop, line 52: Can't we use public-endpoint ?

adityathebe: Isn't that for the incident commander instance and not the frontend URL?

events/notifications.go
```diff
@@ -92,6 +95,57 @@ func (t *NotificationEventProperties) FromMap(m map[string]string) {
...
+		title = "Component {{.component.name}} status updated to {{.component.status}}"
+		body = "Component {{.component.name}} status updated to {{.component.status}}\n\nReference: {{.permalink}}"
+
+	case EventIncidentCommentAdded:
+		title = "New comment on {{.incident.title}}"
+		body = "{{.comment}}\n\nReference: {{.permalink}}"
+
+	case EventIncidentCreated:
+		title = "{{.incident.title}} created"
+		body = "{{.incident.title}} created\n\nReference: {{.permalink}}"
```
moshloop, line 120: Can we add the severity and type and incident id ?

events/notifications.go
```diff
@@ -92,6 +95,57 @@ func (t *NotificationEventProperties) FromMap(m map[string]string) {
...
+
+	case EventIncidentDODAdded:
+		title = "Definition of Done added | {{.incident.title}}"
+		body = "Evidence: {{.evidence.description}}\n\nReference: {{.permalink}}"
```
moshloop, line 124: I think we need to add incident id and title

moshloop:
@adityathebe 
-  Rather than `agent_id:00000000-0000-0000-0000-000000000000 canary_id:018824e5-d0f5-246b-b111-d8ac02c98229` we should have `agent: {id: 000, name: xyz), canary: {id: 000, name: canary, namespace: default}`
- Add error / message to check template
- Add formatted labels 
- Check uptime/latency being 0

events/notifications.go
```diff
@@ -92,6 +96,62 @@ func (t *NotificationEventProperties) FromMap(m map[string]string) {
 	_ = json.Unmarshal(b, &t)
 }
 
+// labelsTemplate is a helper func to generate the template for displaying labels
+func labelsTemplate(field string) string {
+	return fmt.Sprintf("{{if %s}}**Labels**:\n{{range $k, $v := %s}}`{{$k}}={{$v}}`\n{{end}}{{end}}", field, field)
```
moshloop, line 101:
```suggestion
	return fmt.Sprintf("{{if %s}}### Labels\n{{range $k, $v := %s}}`**{{$k}}:** {{$v}}`\n{{end}}{{end}}", field, field)
```

events/notifications.go
```diff
@@ -92,6 +96,62 @@ func (t *NotificationEventProperties) FromMap(m map[string]string) {
...
+// based on the given event.
+func defaultTitleAndBody(event string) (title string, body string) {
+	switch event {
+	case EventCheckPassed:
+		title = "Check {{.check.name}} has passed"
+		body = fmt.Sprintf("%s[Reference]({{.permalink}})", labelsTemplate(".check.labels"))
+
+	case EventCheckFailed:
+		title = "Check {{.check.name}} has failed"
+		body = fmt.Sprintf("%s[Reference]({{.permalink}})", labelsTemplate(".check.labels"))
```
moshloop, line 114: Need to add the check error and message

cmd/root.go
```diff
@@ -48,7 +47,7 @@ func ServerFlags(flags *pflag.FlagSet) {
 	flags.IntVar(&devGuiPort, "devGuiPort", 3004, "Port used by a local npm server in development mode")
 	flags.IntVar(&metricsPort, "metricsPort", 8081, "Port to expose a health dashboard ")
 	flags.BoolVar(&dev, "dev", false, "Run in development mode")
-	flags.StringVar(&publicEndpoint, "public-endpoint", "http://localhost:8080", "Public endpoint that this instance is exposed under")
```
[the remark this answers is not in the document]
adityathebe: It's still there. The new flag is also called `public-endpoint`. I had to move the variable to `api` package

events/notifications.go
adityathebe: @moshloop we either have to do bold or code formatting for the label key

events/notifications.go
```diff
@@ -378,8 +378,14 @@ func getEnvForEvent(ctx *api.Context, eventName string, properties map[string]st
 			check.Latency = summary.Latency
 		}
 
+		var agent models.Agent
+		if err := ctx.DB().Where("id = ?", check.AgentID).First(&agent).Error; err != nil {
```
moshloop, line 382: Can we move the agent, check, canary, person lookups to duty with a cache, e.g. duty.GetCachedAgent() with the specs in check/canary being nulled out to prevent misuse - This cache is only used for immutable (name, namespace, type) and mostly immutable (labels)

adityathebe: How would we keep the cache in sync though?

adityathebe: https://github.com/flanksource/duty/pull/271

events/notifications.go
```diff
@@ -92,11 +98,68 @@ func (t *NotificationEventProperties) FromMap(m map[string]string) {
...
+	return fmt.Sprintf("{{if %s}}### Labels: \n{{range $k, $v := %s}}**{{$k}}**: {{$v}} \n{{end}}{{end}}", field, field)
+}
+
+// defaultTitleAndBody returns the default title and body for notification
+// based on the given event.
+func defaultTitleAndBody(event string) (title string, body string) {
+	switch event {
+	case EventCheckPassed:
+		title = "Check {{.check.name}} has passed"
+		body = fmt.Sprintf("Canary: {{.canary.name}} \nAgent: {{.agent.name}} {{if .check_status.message}}\nMessage: {{.check_status.message}} {{end}}\n%s \n\n[Reference]({{.permalink}})", labelsTemplate(".check.labels"))
```
moshloop, line 112: Only specigy agent if is non zero - can we also use a multiline string instead of a very long line?

events/notifications.go
```diff
@@ -291,34 +358,65 @@ func addNotificationEvent(ctx *api.Context, event api.Event) error {
...
+			return nil, fmt.Errorf("error finding canary: %v", err)
+		} else if canary == nil {
+			return nil, fmt.Errorf("canary(id=%s) not found", check.CanaryID)
+		}
+
+		agent, err := duty.FindCachedAgent(ctx, check.AgentID.String())
+		if err != nil {
+			return nil, fmt.Errorf("error finding agent: %v", err)
+		} else if agent == nil {
+			return nil, fmt.Errorf("agent(id=%s) not found", check.AgentID)
```
moshloop, line 385: this shouldn't be an error condition

events/notifications.go
```diff
@@ -291,34 +358,65 @@ func addNotificationEvent(ctx *api.Context, event api.Event) error {
...
+			check.Latency = summary.Latency
+		}
+
+		// We fetch the latest check_status at the time of event creation
+		var checkStatus models.CheckStatus
+		if err := ctx.DB().Where("check_id = ?", checkID).Where("status = ?", check.Status == models.CheckStatusHealthy).Where("created_at >= ?", event.CreatedAt).Order("created_at").First(&checkStatus).Error; err != nil {
+			return nil, fmt.Errorf("failed to get check status: %w", err)
+		}
+
+		env["check_status"] = checkStatus.AsMap()
```
moshloop, line 402:
```suggestion
		env["status"] = checkStatus.AsMap()
```

PR #548, opened 2023-09-12
fix: support connection lookup by ID

adityathebe:
Related: https://github.com/flanksource/flanksource-ui/issues/1378


moshloop:
Isn't this already done via https://github.com/flanksource/duty/blob/main/connection.go#L67 ? or maybe we just need to handle the `connection://` prefix ?

adityathebe:
@moshloop ya duty already handles it. However, incident commander was returning nil if the connection name didn't have `connection://` prefix.

moshloop:
@adityathebe I think then lets add the support for the prefix in duty

adityathebe:
@moshloop This PR is still needed though. Otherwise, HydrateConnection will never hit duty when a UUID is provided.

adityathebe:
Related: https://github.com/flanksource/duty/pull/275

PR #550, opened 2023-09-13
feat: Notifications CRD

adityathebe:
Resolves: https://github.com/flanksource/incident-commander/issues/547


db/notifications.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"encoding/json"
+
+	"github.com/flanksource/duty/models"
+	v1 "github.com/flanksource/incident-commander/api/v1"
+	"github.com/google/uuid"
+)
+
+func PersistNotificationFromCRD(obj *v1.Notification) error {
```
adityathebe, line 11: @moshloop For all the CRD implemented resources, we have a `spec` column in the corresponding database table. However, notification is already built in a different way. We don't have a `spec` column for it.

Instead of modifying the notification table by removing existing columns and adding the `spec` column, I've implemented it this way. So far I don't really see why this would be an issue.

api/v1/notification_types.go
```diff
@@ -0,0 +1,63 @@
...
+	Person string `json:"person,omitempty" yaml:"person,omitempty"`
+
+	// name or ID of the recipient team
+	Team string `json:"team,omitempty" yaml:"team,omitempty"`
+
+	// Custom services to send this notification to
+	CustomServices []api.NotificationConfig `json:"custom_services,omitempty" yaml:"custom_services,omitempty"`
+
+	// Properties for Shoutrrr
+	Properties map[string]string `json:"properties,omitempty" yaml:"properties,omitempty"`
```
moshloop, line 32: I think the spec should look something like this:
```yaml
to:
   person: 
   team: 
   connection:
   url: 
   properties:
```

I don't think we need to support multiple custom services

db/notifications.go
```diff
@@ -0,0 +1,64 @@
...
+	if obj.Spec.Person != "" {
+		if uid, err := uuid.Parse(obj.Spec.Person); err == nil {
+			dbObj.PersonID = &uid
+		} else {
+			var person models.Person
+			if err := Gorm.Where("email = ?", obj.Spec.Person).First(&person).Error; err != nil {
+				return err
+			}
+			dbObj.PersonID = &person.ID
+		}
```
moshloop, line 35: I think this should move to duty.FindPerson/Team

api/v1/notification_types.go
```diff
@@ -0,0 +1,79 @@
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type NotificationRecipientSpec struct {
```
moshloop, line 7: I think we are missing properties here as well

db/notifications.go
adityathebe: https://github.com/flanksource/duty/pull/277

api/v1/notification_types.go
adityathebe: Why do we need it in here as we already have it in the notification spec?

db/notifications.go
```diff
@@ -0,0 +1,81 @@
...
+	case obj.Spec.To.Person != "":
+		if uid, err := uuid.Parse(obj.Spec.To.Person); err == nil {
+			dbObj.PersonID = &uid
+		} else {
+			var person models.Person
+			if err := Gorm.Where("email = ?", obj.Spec.To.Person).First(&person).Error; err != nil {
+				return err
+			}
+			dbObj.PersonID = &person.ID
+		}
```
moshloop, line 42: Can we move these into a FindPerson/Team?  - I also think the person lookup should be uuid, name or email and our docs to use Name so as not confuse with to.email which doesn't require a person object

adityathebe: >  I also think the person lookup should be uuid, name or email

`people.name` is not unique. Should we add that constraint?

api/v1/notification_types.go
moshloop: It should be in here instead of the notification spec, 

```yaml
to:
    url: smtp://system
    properties:
       FromName: ABC
```

vs which is unclear exactly what the properties apply to

```yaml
to:
    url: smtp://system
properties:
    FromName: ABC
```

PR #554, opened 2023-09-15
feat: single pgNotify listener

adityathebe:
Resolves: https://github.com/flanksource/incident-commander/issues/545
Resolves: https://github.com/flanksource/incident-commander/issues/522


events/eventconsumer/event_consumer.go
```diff
@@ -120,9 +109,6 @@ func (e *EventConsumer) Listen() {
 	// Consume pending events
 	e.ConsumeEventsUntilEmpty(ctx)
 
-	pgNotify := make(chan string)
```
moshloop, line 123: The go routines need to be created here - ConsumeEventsUntilEmpty should be sync and not spawn any go routines

events/pg_notify_router.go
```diff
@@ -0,0 +1,51 @@
...
+	)
+
+	eventQueueNotifyChannel := make(chan string)
+	go utils.ListenToPostgresNotify(t.pgpool, channel, dbReconnectMaxDuration, dbReconnectBackoffBaseDuration, eventQueueNotifyChannel)
+
+	logger.Infof("running pg notify router")
+	for payload := range eventQueueNotifyChannel {
+		if ch, ok := t.registry[payload]; ok {
+			ch <- payload
+		} else if payload != "push_queue.create" { // Ignore push queue events cuz that'll pollute the logs
```
yashmehrotra, line 47: push_queue.create should be this constant: https://github.com/flanksource/incident-commander/blob/main/events/event_queue.go#L79

utils/pg_notify.go
```diff
@@ -24,7 +24,7 @@ func ListenToPostgresNotify(pool *pgxpool.Pool, channel string, dbReconnectMaxDu
 		if err != nil {
```
yashmehrotra, line 24: Shouldn't this file belong in the events package ? can't think of any usecases where it can be used as a generic utility

events/pg_notify_router.go
```diff
@@ -0,0 +1,51 @@
...
+func (t *pgNotifyRouter) Run(channel string) {
+	const (
+		dbReconnectMaxDuration         = time.Minute * 5
+		dbReconnectBackoffBaseDuration = time.Second
+	)
+
+	eventQueueNotifyChannel := make(chan string)
+	go utils.ListenToPostgresNotify(t.pgpool, channel, dbReconnectMaxDuration, dbReconnectBackoffBaseDuration, eventQueueNotifyChannel)
+
+	logger.Infof("running pg notify router")
```
yashmehrotra, line 43: Debugf ?

utils/pg_notify.go
adityathebe: This was moved to `utils` to avoid cyclic dependency.

package `playbook` needs it. And package `events` also needs `playbook`.

events/event_queue.go
```diff
@@ -114,7 +119,8 @@ func StartConsumers(gormDB *gorm.DB, pgpool *pgxpool.Pool, config Config) {
 			uniqEvents[event] = struct{}{}
 		}
 
-		go allSyncHandlers[i].EventConsumer(gormDB, pgpool).Listen()
+		pgNotifyChannel := notifyRouter.RegisterRoutes(allSyncHandlers[i].watchEvents)
```
moshloop, line 122: Is this the pg notify channel or a notification channel

events/eventconsumer/event_consumer.go
```diff
@@ -84,32 +72,22 @@ func (e EventConsumer) Validate() error {
...
-			err := t.consumerFunc(ctx)
-			if err != nil {
-				if api.ErrorCode(err) == api.ENOTFOUND {
-					wg.Done()
-					return
-				}
-
-				logger.Errorf("error processing event, waiting %s to try again: %v", waitDurationOnFailure, err)
-				time.Sleep(waitDurationOnFailure)
+	for {
```
moshloop, line 77: Isn't this an endless loop if there is nothing to consume?

adityathebe: If there's nothing to consuming we return `api.ENOTFOUND` which terminates this loop.

events/event_queue.go
adityathebe: pg notify channel but distributed by the NotifyRouter for the given watch events.

events/eventconsumer/event_consumer.go
moshloop: I don't think we should be using an error for that - in some places we use it when a resource id is not found (an error) and here we are using it for an expected condition - Its probably better to return the number of events consumed

utils/pg_notify.go
adityathebe: @moshloop `pgnotifyrouter` is a very specific use case of PGNotify where we route based on the payload. This model doesn't fit well with all of our uses cases. Example: PGNotify on `playbook_updated` sends the updated playbook id as the payload.

I think, we can move the router to a new `events/pgnotifyrouter` but the `ListenToPostgresNotify` func in utils should stay.

moshloop: @adityathebe  do we not have the same problem for playbook_updated, we could have multiple background workers, but only 1 should receive the event ?

I also don't think utils should ever be used as a final package name - I would prefer moving it to utils/pg/notify.go and then the call becomes `pg.Listen(pool, channel)`  or better yet, `duty/pg/notify.go`

adityathebe: > do we not have the same problem for playbook_updated, we could have multiple background workers, but only 1 should receive the event

We don't. We're still working with a single notify channel & only one of the workers will pick up the signal at a time.

> I also don't think utils should ever be used as a final package name - I would prefer moving it to utils/pg/notify.go and then the call becomes pg.Listen(pool, channel) or better yet, duty/pg/notify.go

Agreed 👍 will move to duty

adityathebe: needed to add `Pool()` to our api.Context struct.
I've made a separate PR that makes this change https://github.com/flanksource/incident-commander/pull/566.

PR #556, opened 2023-09-15
feat: use gomail for shoutrrr SMTP

adityathebe:
Resolves: https://github.com/flanksource/incident-commander/issues/551

### DIFF
--- a/notification/shoutrrr.go
+++ b/notification/shoutrrr.go
@@ -91,6 +91,25 @@ func Send(ctx *api.Context, connectionName, shoutrrrURL, title, message string, 
 	var params *types.Params
 	if properties != nil {
 		params = (*types.Params)(&allProps)
+	} else {
+		params = &types.Params{}
+	}
+
+	// NOTE: Until shoutrrr fixes the "UseHTML" props, we'll use the mailer package
+	if service == "smtp" {
+		parsedURL, err := url.Parse(shoutrrrURL)
+		if err != nil {
+			return fmt.Errorf("failed to parse shoutrrr URL: %w", err)
+		}
+
+		query := parsedURL.Query()
+		var (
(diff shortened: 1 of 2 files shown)


notification/shoutrrr.go
```diff
@@ -91,6 +91,25 @@ func Send(ctx *api.Context, connectionName, shoutrrrURL, title, message string,
 	var params *types.Params
 	if properties != nil {
 		params = (*types.Params)(&allProps)
+	} else {
+		params = &types.Params{}
+	}
+
+	// NOTE: Until shoutrrr fixes the "UseHTML" props, we'll use the mailer package
+	if service == "smtp" {
+		parsedURL, err := url.Parse(shoutrrrURL)
```
moshloop, line 100: I think we need to handle the smtp user/pass/host/port specified in the url  - only smtp://system should default to the environment variables

notification/shoutrrr.go
```diff
@@ -88,11 +89,32 @@ func Send(ctx *api.Context, connectionName, shoutrrrURL, title, message string,
...
 	injectTitleIntoProperties(service, title, allProps)
 
-	var params *types.Params
+	params := &types.Params{}
 	if properties != nil {
 		params = (*types.Params)(&allProps)
 	}
 
+	// NOTE: Until shoutrrr fixes the "UseHTML" props, we'll use the mailer package
+	if service == "smtp" {
```
moshloop, line 98: But now this doesn't work for smtp://system ?

adityathebe: It does. At this point in code, the url is never `smtp://system`. It's already prepopulated. https://github.com/flanksource/incident-commander/blob/cf47ca1327b40638ce8e569691c270d3d6364723/notification/shoutrrr.go#L54-L60

adityathebe:
https://github.com/containrrr/shoutrrr/pull/388

PR #557, opened 2023-09-15
feat: save history of notification sends

adityathebe:
Resolves: https://github.com/flanksource/incident-commander/issues/553

### DIFF
--- a/jobs/jobs.go
+++ b/jobs/jobs.go
@@ -7,7 +7,6 @@ import (
 	"github.com/flanksource/commons/logger"
 	"github.com/flanksource/duty"
 	"github.com/flanksource/incident-commander/api"
-	"github.com/flanksource/incident-commander/db"
 	"github.com/flanksource/incident-commander/responder"
 	"github.com/flanksource/incident-commander/rules"
 	"github.com/flanksource/incident-commander/upstream"
@@ -31,7 +30,7 @@ func ScheduleFunc(schedule string, fn func()) (any, error) {
 	return FuncScheduler.AddFunc(schedule, fn)
 }
 
-func Start() {
+func Start(context duty.DBContext) {
 	// Running first at startup and then with the schedule
 	TeamComponentOwnershipRun()
 	EvaluateEvidenceScripts()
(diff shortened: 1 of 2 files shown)


notification/context.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func (t *Context) LogSourceEvent(event string, resourceID uuid.UUID) {
+	t.log.SourceEvent = event
+	t.log.ResourceID = resourceID
+}
+
+func (t *Context) LogPersonID(id *uuid.UUID) {
+	t.log.PersonID = id
+}
```
moshloop, line 47: I think these should be called "WithError", "WithMessage etc.

utils/utils.go
```diff
@@ -0,0 +1,9 @@
+package utils
+
+import "github.com/flanksource/commons/logger"
+
+func LogIfError(err error, description string) {
```
moshloop, line 5: Can we move this to a log package so that it becomes `log.IfError` and `log.IfErrorf`

jobs/jobs.go
```diff
@@ -63,6 +66,16 @@ func Start() {
 		logger.Errorf("Failed to schedule job for cleaning up event queue table: %v", err)
 	}
 
+	if _, err := ScheduleFunc(CleanupNotificationSendHistorySchedule, func() {
+		if count, err := duty.DeleteNotificationSendHistory(api.NewContext(db.Gorm, nil), 30); err != nil {
```
moshloop, line 70: nit: The logging of the results should probably be in the function itself, and we should be passing through a DbContext rather than db.Gorm

notification/context.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (t *Context) StartLog() error {
+	return db.PersistNotificationSendHistory(t.Context, t.log)
+}
+
+func (t *Context) EndLog() error {
+	return db.PersistNotificationSendHistory(t.Context, t.log.End())
+}
+
```
moshloop, line 31: We should probably just be called t.db.Save directly, no need for an intermediate function

jobs/jobs.go
adityathebe: > The logging of the results should probably be in the function itself

I am not sure if the duty func should be doing the logging.

> we should be passing through a DbContext rather than db.Gorm

All of the jobs use `db.Gorm` directly and there's no DBContext at this stage yet. I guess we could pass gorm directly to the jobs starting function

```
func Start() {}
```

to
```
func Start(db *gorm.DB) {}
```

PR #566, opened 2023-09-20
refactor: make api.Context an interface

adityathebe:
* removed `utils/pg_notify.go`

Merge after: https://github.com/flanksource/incident-commander/pull/554


db/agents.go
```diff
@@ -5,13 +5,13 @@ import (
...
 	"github.com/flanksource/commons/collections"
+	"github.com/flanksource/duty"
 	"github.com/flanksource/duty/models"
-	"github.com/flanksource/incident-commander/api"
 	"github.com/google/uuid"
 	"gorm.io/gorm"
 )
 
-func FindAgent(ctx *api.Context, name string) (*models.Agent, error) {
+func FindAgent(ctx duty.DBContext, name string) (*models.Agent, error) {
```
moshloop, line 14: Why are we switching to duty context, the api.context extends it ?

adityathebe: since all the methods in `db` package just need the database methods from the context I think it's better to use a narrower interface.

adityathebe: There might be a concern of "polluting" all of mission control with duty context. 
I'm also okay with using api.Context over duty.DBContext.

adityathebe: Using `api.Context` since it can additionally have other useful methods for tracing & logging.

PR #570, opened 2023-09-21
chore: always save notification to job history

adityathebe:


notification/expression.go
```diff
@@ -30,38 +30,49 @@ type ExpressionRunner struct {
...
-	jobHistory.AddError(errMsg)
-	if err := db.PersistJobHistory(ctx, jobHistory.End()); err != nil {
-		logger.Errorf("error persisting job history: %v", err)
-	}
-}
-
 // Eval evaluates the given expression into a boolean.
 // The expression should return a boolean value that's supported by strconv.ParseBool.
 func (t ExpressionRunner) Eval(ctx api.Context, expression string) (bool, error) {
+	jobHistory := models.NewJobHistory("NotificationFilterEval", t.ResourceType, t.ResourceID).Start()
```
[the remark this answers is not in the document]
adityathebe: @moshloop made the change. One issue we still have is how a notification can fail for some events but pass for some other event.

Eg: A notification with filter `check.type = 'http'` might listen on events
- `incident.created` 
- `check.passed`

This could be considered an incorrectly setup notification but that's still allowed. If you have no filter in this case, then that's a perfectly working notification.
Maybe in the UI we should not allow mixing events that have different template variables.

notification/expression.go
```diff
@@ -30,38 +33,61 @@ type ExpressionRunner struct {
...
+			}
+		}
+
+		logs.IfError(db.PersistJobHistory(ctx, jobHistory.End()), "error persisting notification filter evaluation job history")
+		expressionResultCache.SetDefault(t.ResourceID, passingCurrently)
+	}()
+
+	result, err := t.eval(ctx, expression)
+	if err != nil {
+		jobHistory.AddError(err.Error())
```
moshloop, line 58: I think this would still create too many events. 

If I have 1 notifications and 10 notifications per minute, it will insert 10 job history errors per minute. 

Why don;t we just put a status and error on the notification itself, if we have an error then we update the table and skip processing the notification entirely, on change of the notification filter we reset the status and error. 

Would probably need a notify to listen to changes on notifications

events/notifications.go
```diff
@@ -296,11 +288,9 @@ func addNotificationEvent(ctx api.Context, event api.Event) error {
 			CelEnv:       celEnv,
 		}
 
-		if valid, err := expressionRunner.Eval(ctx, n.Filter); err != nil || !valid {
-			// We consider an error in filter evaluation is a failed filter check.
-			// Mostly, the filter check returns an error if the variable isn't defined.
-			// Example: If the filter makes use of `check` variable but the event is for
-			// incident creation, then the expression evaluation returns an error.
+		if valid, err := expressionRunner.Eval(ctx, n.Filter); err != nil {
+			logs.IfError(db.UpdateNotificationError(id, err.Error()), "failed to update notification")
```
moshloop, line 292: I don't see how this prevents the filter from being run again, and will cause the row to be continually be touched?

We should have a StopNotification call here - 

I also think most of these logic should move to the notifications package - the events package should just be handing it off with the environment. 

Why do we have a pkgNotification.getNotificationIds and a pkgNotification.GetNotification(id) - Why not just combine it?

adityathebe: > I don't see how this prevents the filter from being run again, and will cause the row to be continually be touched?
> We should have a StopNotification call here -

The expression is not run if the notification is in error status (line 281)

> Why do we have a pkgNotification.getNotificationIds and a pkgNotification.GetNotification(id) - Why not just combine it?

One's fetching all notifications that are listening on an event and the other one is fetching by given ids

> Also not sure why we have pkgNotification and pkgResponder, why not just use notification?

I think `responder` & `notification` are already used as variables so the package name was prefixed with `pkg`

adityathebe:
@moshloop On a second look, I think what we currently have in `events/notifications.go` file should stay there and shouldn't be move the `notification` package.

I see the `notification` package as our `mail` package which is just the client to send notification. It can be called from any other places to send a notification. So, I think it shouldn't hold in any logic for event queue handling like 
- creating a `notification.send` event from `check.failed` event, or
- deciding the title and body for incident related events.

We could move to a new `notification/events` package instead but I still think what we currently have is fine.

moshloop:
@adityathebe - notifications.go is still too large

- These all should be moved to notification package and the prefix removed: NotificationEventPayload, NotificationTemplate, labelsTemplate, defaultTitleAndBody, sendNotification 
- addNotificationEvent should be trimmed down to only forwarding the relevant details to notification package - e.g. GetNotifications should only be called from within the notification package, 
- getEnvForEvent should remain in events.go

adityathebe:
@moshloop I've made the changes.

PR #575, opened 2023-09-27
feat: add tracing via opentelemetry

yashmehrotra:


events/event_queue.go
```diff
@@ -195,6 +197,11 @@ func (t SyncEventConsumer) EventConsumer() *eventconsumer.EventConsumer {
 }
 
 func (t *SyncEventConsumer) Handle(ctx api.Context) (int, error) {
+	tracer := otel.GetTracerProvider().Tracer("event-tracer")
+	traceCtx, span := tracer.Start(ctx, "event-queue-"+strings.Join(t.watchEvents, "/"))
+	ctx = ctx.WithContext(traceCtx)
+	defer span.End()
+
```
adityathebe, line 204: I think this could be 

```go
ctx, span := ctx.Span(tracer, "event-queue-"+strings.Join(t.watchEvents, "/"))
defer span.End()

```
Where `tracer := otel.GetTracerProvider().Tracer("event-tracer")` is a package level variable.

cmd/server.go
```diff
@@ -128,6 +133,19 @@ func createHTTPServer(ctx api.Context) *echo.Echo {
 	e.Use(middleware.Logger())
 	e.Use(ServerCache)
 
+	// Set UserID in trace headers
+	e.Use(func(next echo.HandlerFunc) echo.HandlerFunc {
+		return func(c echo.Context) error {
+			if userID := c.Request().Header.Get(api.UserIDHeaderKey); userID != "" {
+				span := trace.SpanFromContext(c.Request().Context())
+				span.SetAttributes(
+					attribute.String("user-id", userID),
```
moshloop, line 142: I think we should add org and clerk id in here as well

events/event_queue.go
```diff
@@ -255,6 +262,11 @@ type AsyncEventConsumer struct {
 }
 
 func (t *AsyncEventConsumer) Handle(ctx api.Context) (int, error) {
+	tracer := otel.GetTracerProvider().Tracer("event-tracer")
+	traceCtx, span := tracer.Start(ctx, "event-queue-"+strings.Join(t.watchEvents, "/"))
```
moshloop, line 266: I think the watchEvents should be attributes

playbook/run_consumer.go
```diff
@@ -19,6 +20,11 @@ func StartPlaybookRunConsumer(ctx api.Context) {
 }
 
 func EventConsumer(ctx api.Context) (int, error) {
+	tracer := otel.GetTracerProvider().Tracer("event-tracer")
```
moshloop, line 23: I think this will produce too many background traces - Maybe we can sample this at @ 1% ?

cmd/server.go
```diff
@@ -54,6 +57,8 @@ func createHTTPServer(ctx api.Context) *echo.Echo {
 	e := echo.New()
 	e.HideBanner = true
 
+	e.Use(otelecho.Middleware("mission-control"))
```
moshloop, line 60: I think we should probably gate this and the db middleware if teacing is enabled only

yashmehrotra: It works silently by using noop spans

playbook/run_consumer.go
yashmehrotra: Done

events/event_queue.go
yashmehrotra: 👍🏽

cmd/server.go
yashmehrotra: 👍🏽

events/event_queue.go
yashmehrotra: Makes sense, added a `ctx.StartTrace()` function

auth/clerk_client.go
```diff
@@ -86,6 +88,11 @@ func (h ClerkHandler) Session(next echo.HandlerFunc) echo.HandlerFunc {
 		c.Request().Header.Set(echo.HeaderAuthorization, fmt.Sprintf("Bearer %s", token))
 		c.Request().Header.Set(api.UserIDHeaderKey, user.ID.String())
 
+		trace.SpanFromContext(c.Request().Context()).SetAttributes(
```
moshloop, line 91: Why not just add a ctx.WithAttribute[s] function  and/or update WithUser to set the the trace attributes as well

api/context.go
```diff
@@ -90,6 +96,27 @@ func (c context) WithContext(ctx gocontext.Context) Context {
...
+		return true
+	}
+	if perc <= 0 {
+		return false
+	}
+	randomNum := rand.Intn(100)
+	return perc >= randomNum
+}
+
+func (c context) StartTrace(tracerName, spanName string, samplingPerc int) (Context, trace.Span) {
```
moshloop, line 110: I think it would be better if we split this into a StartTrace and StartSampledTrace

also why do we need both tracerName and span name ?

yashmehrotra: tracer-name and span-name are different tags that otel sets on its own

auth/clerk_client.go
yashmehrotra: 👍🏽

PR #581, opened 2023-10-04
feat: add a max limit on the number of rows that postgrest can return

adityathebe:
https://github.com/flanksource/duty/issues/286

### DIFF
--- a/db/postgrest.go
+++ b/db/postgrest.go
@@ -5,14 +5,14 @@ import (
 	"github.com/flanksource/commons/logger"
 )
 
-var PostgRESTVersion = "v10.0.0"
-var PostgRESTJWTSecret string
-var PostgresDBAnonRole string
+var (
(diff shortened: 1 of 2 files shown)


adityathebe:
I think it would have been better to make the query fail if the rows returned > limit. 
With this setting, it just silently truncates the rows & I couldn't find such setting.

db/postgrest.go
```diff
@@ -9,6 +9,12 @@ var PostgRESTVersion = "v10.0.0"
 var PostgRESTJWTSecret string
 var PostgresDBAnonRole string
 
+const (
+	// A hard limit to the number of rows PostgREST will fetch from a view, table, or stored procedure.
+	// Limits payload size for accidental or malicious requests.
+	postgrestMaxRows = "2000"
```
moshloop, line 15: I think this will need to be configurable

PR #586, opened 2023-10-09
feat: setup otel propagator & service name

adityathebe:
* use new http client from commons
* use upstream client from duty
* use the new `is_pushed` column to sync check statuses

Merge after:
- [x] https://github.com/flanksource/duty/pull/296
- [x] https://github.com/flanksource/commons/pull/79
- [x] https://github.com/flanksource/duty/pull/300

Resolves:
- [x] https://github.com/flanksource/canary-checker/issues/1310


jobs/upstream.go
```diff
@@ -43,8 +42,19 @@ func SyncWithUpstream(ctx api.Context) error {
...
+func reconcileTable(ctx api.Context, table string) error {
+	newCtx, span := tracer.Start(ctx, fmt.Sprintf("reconcile-%s", table))
+	defer span.End()
+
+	ctx = ctx.WithContext(newCtx)
+	reconciler := upstream.NewUpstreamReconciler(api.UpstreamConf, ReconcilePageSize)
+
+	return reconciler.SyncAfter(ctx, table, ReconcileMaxAge)
+}
+
```
adityathebe, line 54: Added a new span for each table we're reconciling because I think it gives a better visual to the trace

image

adityathebe: vs 
image

upstream/controllers.go
```diff
@@ -136,6 +145,8 @@ func PullCanaries(c echo.Context) error {
 		if err != nil {
 			return c.JSON(http.StatusBadRequest, api.HTTPError{Error: err.Error(), Message: "'since' param needs to be a valid RFC3339 timestamp"})
 		}
+
+		ctx.SetSpanAttributes(attribute.String("since", sinceRaw))
```
yashmehrotra, line 149: prefix this with `upstream.since` since these attributes will be at global scope

adityathebe:
@moshloop This is ready to go

PR #587, opened 2023-10-11
feat: add more playbook actions

adityathebe:
Resolves: https://github.com/flanksource/incident-commander/issues/582


api/v1/playbook_actions.go
```diff
@@ -10,6 +10,53 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
+type SQLAction struct {
+	// Connection identifier e.g. connection://Postgres/flanksource
+	Connection string `yaml:"connection,omitempty" json:"connection,omitempty"`
+	// URL is the database connection url
+	URL string `yaml:"url,omitempty" json:"url,omitempty"`
+	// Database is the name of the database
+	Database string `yaml:"database" json:"database"`
```
moshloop, line 19: Isn't this normally part of the URL ?

api/v1/playbook_actions.go
```diff
@@ -10,6 +10,53 @@ import (
...
+	Password types.EnvVar `yaml:"password,omitempty" json:"password,omitempty"`
+}
+
+type HTTPAction struct {
+	HTTPConnection `yaml:",inline" json:",inline"`
+	// Maximum duration in milliseconds for the HTTP request. It will fail the check if it takes longer.
+	ThresholdMillis int `yaml:"thresholdMillis,omitempty" json:"thresholdMillis,omitempty"`
+	// Maximum number of days until the SSL Certificate expires.
+	MaxSSLExpiry int `yaml:"maxSSLExpiry,omitempty" json:"maxSSLExpiry,omitempty"`
+	// Method to use - defaults to GET
```
moshloop, line 46: These aren't required as they are specific to canaries
```suggestion
```

go.mod
```diff
@@ -41,20 +41,23 @@ require (
 )
 
 require (
+	github.com/flanksource/kommons v0.31.4
```
moshloop, line 44: Can we copy and paste the relevant code from kommons, we are trying to deprecate kommons

playbook/actions/pod.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/flanksource/incident-commander/api"
+	v1 "github.com/flanksource/incident-commander/api/v1"
+)
+
+const (
+	defaultMaxLength        = 3000
+	defaultContainerTimeout = 5 * time.Minute
+
+	sidecarContainerName     = "container-waiter"
+	containerImage           = "ubuntu:jammy"
```
moshloop, line 23: I don't think these should have any defaults

playbook/actions/pod.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func (c *Pod) Run(ctx api.Context, action v1.PodAction, env TemplateEnv) (*PodResult, error) {
+	timeout := time.Duration(action.Timeout) * time.Minute
+	if timeout == 0 {
+		timeout = defaultContainerTimeout
+	}
+
+	if action.MaxLength <= 0 {
+		action.MaxLength = defaultMaxLength
```
moshloop, line 42: I don't think we should trim by default - its good to add an option, but not as a default - Long term we would store there logs in S3 like store

playbook/actions/http.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/flanksource/gomplate/v3"
+	"github.com/flanksource/incident-commander/api"
+	v1 "github.com/flanksource/incident-commander/api/v1"
+)
+
+type HTTPResult struct {
+	Code    string
+	Headers netHTTP.Header
+	Body    string
+	SslAge  *time.Duration
```
moshloop, line 20: I don't think this is needed
```suggestion
```

playbook/actions/http.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+type HTTP struct {
+}
+
+func (c *HTTP) Run(ctx api.Context, action v1.HTTPAction, env TemplateEnv) (*HTTPResult, error) {
+	if action.URL == "" {
```
moshloop, line 27: Couldn't the URL come from a connection

playbook/actions/pod.go
```diff
@@ -0,0 +1,116 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/rand"
+
+	"github.com/flanksource/incident-commander/api"
+	v1 "github.com/flanksource/incident-commander/api/v1"
+)
+
+const (
+	defaultMaxLength        = 3000
+	defaultContainerTimeout = 5 * time.Minute
```
moshloop, line 20:
```suggestion
	defaultContainerTimeout = 30 * time.Minute
```

playbook/actions/pod.go
```diff
@@ -0,0 +1,116 @@
...
+	if action.MaxLength <= 0 {
+		action.MaxLength = defaultMaxLength
+	}
+
+	pod, err := newPod(ctx, action)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := cleanupExistingPods(ctx, fmt.Sprintf("%s=%s", playbookActionLabelValue, pod.Labels[playbookActionLabelValue])); err != nil {
```
moshloop, line 50: I am not sure we want to cleanup existing pods, during a playbook execution - Considering we may run the same step many times for different inputs. 

The pod name should probably be generated from the run_step_id, and then get labelled with the playbook name, id, stepname and step id and then have seperate job to cleanup orphans - maybe daily

PR #601, opened 2023-10-16
feat: auto approve runs & add timeout to actions

adityathebe:
resolves: https://github.com/flanksource/incident-commander/issues/519


playbook/controllers.go
```diff
@@ -143,6 +144,22 @@ func HandlePlaybookRun(c echo.Context) error {
 		return c.JSON(http.StatusInternalServerError, api.HTTPError{Error: err.Error(), Message: "failed to create playbook run"})
 	}
 
+	go func() {
+		// Create a new context because the original request context will be cancelled immediately
```
moshloop, line 148: Why does the context need to remain active after approving a run ? The run should be picked up in the background, and not synchronously during this call

playbook/controllers.go
```diff
@@ -143,6 +144,22 @@ func HandlePlaybookRun(c echo.Context) error {
...
+		// after exiting out of this handler.
+		ctx, cancel := api.NewContext(ctx.DB(), ctx.Pool()).WithUser(ctx.User()).WithTimeout(time.Second * 20)
```
moshloop, line 150: This type of new decentralised context creation / cloning  I want to avoid - If we need a context that lives past the request then it should be a new context function that passes in relevant details
 e.g. 

For example we wanted to enable tracing by setting a trace header or request, it would be done in the api.NewContextFromHttpRequest(ctx) or ctx.NewBackgroundContext()

   or` api.NewContextFromHttpRequest() { api.DefaultContext.WithTrace(req.headers['trace'] = 'true').WithValue("startedBy", user.id|)`

playbook/controllers.go
adityathebe: It doesn't remain active after approving a run?

It's also run asynchronously.

playbook/controllers.go
adityathebe: I've added a new `WithDetachedTimeout` to the context which essentially preserves everything in the context apart from the timeout.

https://github.com/flanksource/incident-commander/blob/eaacae2e2b3d98fe5974e0626579e15a9ff9ca70/api/context_test.go

adityathebe: It uses a new function from the context package introduced in go 1.21

playbook/controllers.go
moshloop: It shouldn't be be kicked off by the http request - the http request updates the db and then the normal notify mechanism's should schedule it to run - the scheduling is pretty simple now, but we could have dedicated runners, pools of runners, runners with specific tags, etc..

api/context.go
```diff
@@ -105,12 +107,18 @@ func (c apicontext) WithContext(ctx gocontext.Context) Context {
...
 }
 
-func (c apicontext) WithTimeout(timeout time.Duration) (Context, func()) {
+func (c apicontext) WithTimeout(timeout time.Duration) (Context, gocontext.CancelFunc) {
 	ctx, cancel := gocontext.WithTimeout(c.Context, timeout)
 	c.Context = ctx
 	return &c, cancel
 }
 
+func (c apicontext) WithDetatchedTimeout(timeout time.Duration) (Context, gocontext.CancelFunc) {
```
moshloop, line 116: Can we remove this - I would prefer to only start using 1.21 API's when we know its working afterting deploying to demo etc..

PR #608, opened 2023-10-18
refactor: depreate api.Context for playbook and upstream

yashmehrotra:


playbook/controllers.go
```diff
@@ -200,7 +201,7 @@ func HandlePlaybookList(c echo.Context) error {
 }
 
 func HandlePlaybookRunApproval(c echo.Context) error {
-	ctx := c.(api.Context)
+	ctx := c.Request().Context().(context.Context)
```
moshloop, line 204: Weren't we going to wrap the echo context ?

yashmehrotra: We are wrapping it in the middleware: https://github.com/flanksource/incident-commander/blob/main/cmd/server.go#L77

Just need to type-assert in the funcs, or should I replace this with wrap as well ?

yashmehrotra:
```go
ctx := api.DefaultContext.Wrap(c.Request().Context())

// or

ctx := c.Request().Context().(context.Context)
```

yashmehrotra: actually, we have to do it in the middleware itself cause the auth middleware needs duty.Context

adityathebe:
@yashmehrotra I think you've covered almost all controllers _(apart from maybe just the agent's controller)_. Why not get rid of `api.Context` totally from the middleware in this PR itself?

`func createHTTPServer` would accept duty's context in that case.

playbook/controllers.go
adityathebe: > We are wrapping it in the middleware: https://github.com/flanksource/incident-commander/blob/main/cmd/server.go#L77
> 
> Just need to type-assert in the funcs, or should I replace this with wrap as well ?

The way we are wrapping contexts in middleware is very confusing

`c.SetRequest(c.Request().WithContext(api.DefaultContext.Wrap(c.Request().Context())))`

yashmehrotra:
> @yashmehrotra I think you've covered almost all controllers (apart from maybe just the agent's controller). Why not get rid of api.Context totally from the middleware in this PR itself?

@adityathebe Doing that in a different PR with other changes, wanted to keep the scope smaller for this one

playbook/controllers.go
yashmehrotra: Should we create a func around it ?
```go
func WithEchoContext(c echo.Context, ctx dutycontext.Context) *http.Request {

    return c.Request().WithContext(
      ctx.Wrap(c.Request().Context())
  )
  ```
  
  And do  this in middleware ?
  ```go

c.SetRequest(WithEchoContext(c, api.DefaultContext))
```

adityathebe: @yashmehrotra Not sure. We're still doing the same thing but in a different place. But also there doesn't seem to be a cleaner way to inject our context into echo's context.

We also could use `c.Set("duty-context", api.DefaultContext.Wrap(c.Request().Context())` although maybe that's not any better.

yashmehrotra: > But also there doesn't seem to be a cleaner way to inject our context into echo's context.
yeah, i think lets leave it to original approach as its explicit

PR #622, opened 2023-10-30
fix: timestamps for playbooks

adityathebe:
resolves: https://github.com/flanksource/incident-commander/issues/621


adityathebe:
image

playbook/run_consumer.go
```diff
@@ -51,6 +51,7 @@ func EventConsumer(c postq.Context) (int, error) {
 			AND scheduled_time <= NOW()
 		ORDER BY scheduled_time
 		FOR UPDATE SKIP LOCKED
+		LIMIT 5
```
adityathebe, line 54: Added a limit here so a single consumer doesn't keep hold of many playbook runs.

PR #627, opened 2023-11-01
feat: mapping identities to roles & teams

adityathebe:
- [x] The argument should also accept a path to read the script, which we mount into a config map from the chart
- [x] the evaluated role from the expression should be persisted

### DIFF
--- a/auth/middleware.go
+++ b/auth/middleware.go
@@ -42,6 +42,11 @@ var (
 	identityMapperLoginCache = cache.New(1*time.Hour, 1*time.Hour)
 )
 
+type IdentityMapperExprResult struct {
+	Teams []string `json:"teams"`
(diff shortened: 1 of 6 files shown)


auth/middleware.go
```diff
@@ -103,13 +116,51 @@ func (k *kratosMiddleware) Session(next echo.HandlerFunc) echo.HandlerFunc {
...
+				var result map[string]string
+				if err := json.Unmarshal([]byte(res), &result); err != nil {
+					return err
+				}
+
+				if res, ok := result["role"]; ok {
+					ctx.Context = ctx.WithValue("identity.role", res)
+				}
+
+				if team, ok := result["team"]; ok {
```
moshloop, line 142: This should support teams[]

PR #645, opened 2023-11-15
chore: default clerk role to viewer

yashmehrotra:
Fixes: #644


moshloop:
@yashmehrotra  - I experienced this when accepting the invite for cats, and upon login got lots of permission denied errors.

PR #652, opened 2023-11-23
feat: gitops action

adityathebe:
resolves: https://github.com/flanksource/incident-commander/issues/490


playbook/actions/gitops.go
```diff
@@ -0,0 +1,236 @@
...
+			return err
+		}
+
+		switch conn.Type {
+		case models.ConnectionTypeGithub:
+			t.spec.GITHUB_TOKEN = conn.Password
+			t.shouldCreatePR = true
+
+		case models.ConnectionTypeAzureDevops:
+			// TODO: Azure devops connection doesn't have git credentials ...?
```
moshloop, line 119: The token would be the http password

playbook/actions/gitops.go
```diff
@@ -0,0 +1,236 @@
...
+
+		case models.ConnectionTypeGit:
```
moshloop, line 121: the would be case git or http

Bot http and  git connection can specify a username and password, and git can specify a SSH clone key (certificate)

playbook/actions/gitops.go
```diff
@@ -0,0 +1,236 @@
...
+			// TODO: need to finalize this once git connection is implemented
+
+		default:
+			return fmt.Errorf("unsupported connection type: %s", conn.Type)
+		}
+	}
+
+	if action.PullRequest == nil {
+		logger.Warnf("no pull request spec was provided on gitops action. So no PR will be created")
```
moshloop, line 130:
```suggestion
```

playbook/actions/gitops.go
```diff
@@ -0,0 +1,236 @@
...
+
+		fullpath := filepath.Join(t.workTree.Filesystem.Root(), patch.Path)
+		if patch.YQ != "" {
+			script, err := gomplate.RunTemplate(t.env.AsMap(), gomplate.Template{Template: patch.YQ})
+			if err != nil {
+				return err
+			}
+			patch.YQ = script
+
+			cmd := exec.Command("yq", "eval", "-i", patch.YQ, fullpath)
```
moshloop, line 180: We need to capture the output to display in results

playbook/actions/gitops.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+func (t *GitOps) applyPatches(ctx context.Context, action v1.GitOpsAction) error {
+	for _, patch := range action.Patches {
+		path, err := gomplate.RunTemplate(t.env.AsMap(), gomplate.Template{Template: patch.Path})
+		if err != nil {
+			return err
+		}
+
+		fullpath := filepath.Join(t.workTree.Filesystem.Root(), patch.Path)
```
moshloop, line 172: Add a log line to add results.stdout: 'Patching {file}' - I think we should also unglob

playbook/actions/gitops.go
```diff
@@ -0,0 +1,236 @@
...
+
+	return nil
+}
+
+func (t *GitOps) modifyFiles(ctx context.Context, action v1.GitOpsAction) error {
+	var err error
+
+	for _, files := range action.Files {
+		for path, content := range files {
+			path, err = gomplate.RunTemplate(t.env.AsMap(), gomplate.Template{Template: path})
```
moshloop, line 203: Unglob + and add log messages

playbook/actions/gitops.go
```diff
@@ -0,0 +1,236 @@
...
+
+	return &response, nil
+}
+
+// generateSpec generates the spec for the git client from the action
+func (t *GitOps) generateSpec(ctx context.Context, action v1.GitOpsAction) error {
+	var err error
+
+	if action.Repo.Base == "" {
+		action.Repo.Base = "master"
```
moshloop, line 79:
```suggestion
		action.Repo.Base = "main"
```

playbook/actions/gitops.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+
+	if action.Repo.Branch == "" {
```
moshloop, line 82: I think we can just use gomplate.StructWalker to template all repo values

playbook/actions/gitops.go
```diff
@@ -0,0 +1,236 @@
+package actions
```
moshloop, line 1: Can we add tests using https://github.com/fluxcd/pkg/tree/main/gittestserver ?

playbook/actions/gitops.go
```diff
@@ -0,0 +1,236 @@
...
+
+type GitOpsActionResult struct {
+	CreatedPR int `json:"createdPR,omitempty"`
+}
+
+func (t *GitOps) Run(ctx context.Context, action v1.GitOpsAction, env TemplateEnv) (*GitOpsActionResult, error) {
+	var response GitOpsActionResult
+
+	if len(action.Patches) == 0 && len(action.Files) == 0 {
+		logger.Warnf("no patches or files specified on gitops action")
```
moshloop, line 35: This should be an error

playbook/actions/gitops.go
```diff
@@ -0,0 +1,236 @@
...
+	if action.Repo.Branch == "" {
+		action.Repo.Branch = action.Repo.Base
+	} else {
+		action.Repo.Branch, err = gomplate.RunTemplate(t.env.AsMap(), gomplate.Template{Template: action.Repo.Branch})
+		if err != nil {
+			return err
+		}
+	}
+
+	t.spec = &connectors.GitopsAPISpec{
```
moshloop, line 91: Why do we need GitopsAPI Spec ?

adityathebe: This came directly from tenant-controller. And I tried to use it as it is an external lib and then convert our spec to `GitopsAPISpec` . But I guess we could just directly use `v1.GitOpsAction`

fixtures/playbooks/gitops.yaml
```diff
@@ -0,0 +1,42 @@
...
+            yq: '.metadata.namespace = "{{.params.namespace}}"'
+        files:
+          - path: fixtures/topology/single-check.yaml
+            content: $delete
+          - path: '{{.params.namespace}}-ns.yaml'
+            content: |
+              apiVersion: v1
+              kind: Namespace
+              metadata:
+                name:  {{.params.namespace}}
```
adityathebe, line 42: @moshloop `files` couldn't just be `map[string]string` as I think the order matters here.

With it changed to `[]map[string]string`, struct templater couldn't template it. 
So, rather I changed to this format.

playbook/actions/gitops.go
```diff
@@ -0,0 +1,232 @@
...
+		if err != nil {
+			return err
+		}
+
+		for _, path := range paths {
+			relativePath, err := filepath.Rel(t.workTree.Filesystem.Root(), path)
+			if err != nil {
+				return fmt.Errorf("failed to get relative path: %w", err)
+			}
+			logger.Debugf("Patching %s", relativePath)
```
moshloop, line 171: These logs should be GitOps object/context so that they can be saved to the db and shown in the ui

playbook/actions/gitops.go
adityathebe: This command doesn't return any output and modifies the file in-place.

We could just read the file content if it's needed?

PR #664, opened 2023-12-01
feat: add git checkout to exec action

adityathebe:
resolves: https://github.com/flanksource/incident-commander/issues/658


moshloop:
@adityathebe can you regen ?

Issue #790, opened 2024-01-26
Handle saving artifacts on the agents


adityathebe:
The agents do not have the credentials for the artifact store; only the main instance does.
So, the agents should use a (customizable) local fs as the artifact store. And then the artifacts will be pushed to the upstream where the upstream will put those artifacts in the actual artifact store.

PR #793, opened 2024-01-29
feat: push agent artifacts

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/790

merge after: https://github.com/flanksource/duty/pull/475


jobs/upstream.go
```diff
@@ -89,3 +89,43 @@ var SyncArtifactRecords = &job.Job{
...
+// agentArtifactPath is the local path to the agent artifact store.
+var agentArtifactPath string
+
+// SyncArtifactRecords pushes any unpushed artifact records to the upstream.
+// The actual artifacts aren't pushed by this job.
+var SyncArtifactData = &job.Job{
+	JobHistory: true,
+	Singleton:  false, // this job is safe to run concurrently
+	RunNow:     true,
+	Retention:  job.RetentionHour,
```
moshloop, line 102: This is too high of a retention for a job that runs every 30s

jobs/upstream.go
```diff
@@ -89,3 +89,43 @@ var SyncArtifactRecords = &job.Job{
...
+	Name:       "SyncArtifactData",
+	Schedule:   "@every 30s",
+	Fn: func(ctx job.JobRuntime) error {
+		ctx.History.ResourceType = job.ResourceTypeUpstream
+		ctx.History.ResourceID = api.UpstreamConf.Host
+
+		if agentArtifactPath == "" {
```
moshloop, line 109: Why can't we use the default artifact store and then support an inline definition of the connection e.g. `fs://var/artifacts`

artifacts/artifacts.go
```diff
@@ -0,0 +1,47 @@
...
+	dutyAPI "github.com/flanksource/duty/api"
+	"github.com/flanksource/duty/context"
+	"github.com/flanksource/duty/models"
+	"github.com/flanksource/incident-commander/api"
+	"github.com/google/uuid"
+	"gorm.io/gorm"
+)
+
+// UploadArtifact Uploads the given artifact data to the default artifact store
+func UploadArtifact(ctx context.Context, artifactID string, reader io.ReadCloser) error {
```
moshloop, line 18: I am a little confused by this function - Don't we upload to the artifact store and then insert the db record ?

adityathebe: It's the other way around - the artifact db records are first pushed and then the data

jobs/upstream.go
adityathebe: But we are using the default artifact store. 

An inline definition would not work because saving the artifact on the agent would require a connection id.
We can make it work by going around the foreign key constraint but I don't see the point.

In the agent chart, we can have a default artifacts connection that's of type 'folder' with path '/var/artifacts'

playbook/run_consumer.go
```diff
@@ -83,9 +83,16 @@ func ActionConsumer(c postq.Context) (int, error) {
 	}
 
 	for i := range foundActions {
+		isAssignedToAgent := foundActions[i].AgentID != nil && *foundActions[i].AgentID != uuid.Nil
```
adityathebe, line 86: This fixes a bug where the host runs an action destined for an agent.

playbook/run_consumer.go
```diff
@@ -185,16 +192,14 @@ func RunConsumer(c postq.Context) (int, error) {
 		SELECT playbook_runs.*
 		FROM playbook_runs
 		INNER JOIN playbooks ON playbooks.id = playbook_runs.playbook_id
-		WHERE status = ?
```
adityathebe, line 188: This fixes a bug where a playbook with only agent runners would never be marked as completed

PR #824, opened 2024-02-13
feat: azure devops pipeline playbook action

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/733


api/v1/playbook_actions.go
```diff
@@ -98,6 +98,26 @@ type GithubAction struct {
...
+
+type AzureDevopsAction struct {
+	// Org is the name of the Azure DevOps organization
+	Org string `json:"org" template:"true"`
+	// Project ID or project name
+	Project string `json:"project" template:"true"`
+	// Token is the personal access token
+	Token types.EnvVar `yaml:"token" json:"token"`
+	// Pipelines is the list of azure pipelines to invoke
+	Pipelines []AzureDevopsPipeline `json:"pipelines,omitempty" template:"true"`
```
moshloop, line 118: I don't think we would execute multiple pipelines in 1 action, 1 action == 1 pipeline run so we can inline them here:

The action would take an id for pipeline id, branch and a map of parameters.

api/v1/playbook_actions.go
```diff
@@ -362,6 +382,7 @@ type PlaybookAction struct {
 	Exec         *ExecAction         `json:"exec,omitempty" yaml:"exec,omitempty" template:"true"`
 	GitOps       *GitOpsAction       `json:"gitops,omitempty" yaml:"gitops,omitempty" template:"true"`
 	Github       *GithubAction       `json:"github,omitempty" yaml:"github,omitempty" template:"true"`
+	AzureDevops  *AzureDevopsAction  `json:"azureDevops,omitempty" yaml:"azureDevops,omitempty" template:"true"`
```
moshloop, line 385: This should probably be azureDevopsPipeline

adityathebe:
image

api/v1/playbook_actions.go
```diff
@@ -98,6 +98,54 @@ type GithubAction struct {
...
+
+	StagesToSkip []string `json:"stagesToSkip,omitempty" template:"true"`
+
+	// If true, don't actually create a new run.
+	// Instead, return the final YAML document after parsing templates.
+	PreviewRun bool `json:"previewRun,omitempty" template:"true"`
+
+	// If you use the preview run option, you may optionally supply different YAML.
+	// This allows you to preview the final YAML document without committing a changed file.
+	YamlOverride string `json:"yamlOverride,omitempty" template:"true"`
```
moshloop, line 133: I don't think using these options would make sense.

api/v1/playbook_actions.go
```diff
@@ -98,6 +98,54 @@ type GithubAction struct {
 	Workflows []GithubWorkflow `yaml:"workflows,omitempty" json:"workflows,omitempty" template:"true"`
 }
 
+type AzureDevopsPipeline struct {
+	// ID is the pipeline ID
+	ID string `json:"id" template:"true"`
+	// Version is the pipeline version
+	Version string `json:"version,omitempty" template:"true"`
+	// ApiVersion is the version of the API to use. Defaults to "7.1-preview.1"
+	ApiVersion string `json:"apiVersion,omitempty" template:"true"`
```
moshloop, line 107:
```suggestion
```

PR #848, opened 2024-02-26
feat: add playbook spec validation

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/845


config/schemas/schemas.go
```diff
@@ -0,0 +1,6 @@
+package schemas
```
moshloop, line 1: Do we need this anymore?

PR #855, opened 2024-02-28
feat: add playbook filter

yashmehrotra:
Depends on: https://github.com/flanksource/duty/pull/551


playbook/runner.go
```diff
@@ -168,6 +168,32 @@ func GetActionForAgent(ctx context.Context, agent *models.Agent) (*ActionForAgen
...
+		return false, err
+	}
+	for _, f := range playbookSpec.Filters {
+		val, err := ctx.RunTemplate(gomplate.Template{Expression: f}, templateEnv.AsMap())
+		if err != nil {
+			ctx.Errorf("error running playbook[%s] filter expression[%s]: %v", playbook.ID, f, err)
+			return false, nil
+		}
+
+		// The expression must return a boolean
```
moshloop, line 189: I think it should support a string return which would be used as the error message

playbook/runner.go
```diff
@@ -179,6 +205,15 @@ func HandleRun(ctx context.Context, run models.PlaybookRun) error {
 		return fmt.Errorf("failed to fetch playbook: %w", err)
 	}
 
+	filterPassed, err := checkPlaybookFilter(ctx, playbook, run)
+	if err != nil {
+		return fmt.Errorf("error running playbook filter check: %w", err)
+	}
+	if !filterPassed {
+		// TODO: Create Playbook Status in duty
+		return ctx.DB().Model(&models.PlaybookRun{}).Where("id = ?", run.ID).UpdateColumn("status", "filter failed").Error
```
moshloop, line 214:
```suggestion
		return ctx.DB().Model(&models.PlaybookRun{}).Where("id = ?", run.ID).UpdateColumn("status", "Skipped").Error
```

playbook/runner.go
```diff
@@ -179,6 +205,15 @@ func HandleRun(ctx context.Context, run models.PlaybookRun) error {
 		return fmt.Errorf("failed to fetch playbook: %w", err)
 	}
 
+	filterPassed, err := checkPlaybookFilter(ctx, playbook, run)
```
moshloop, line 208: This check should be done in the parameter templating function so we don't event get to execution

playbook/runner.go
yashmehrotra: Where will we propagate the message ? should we add a details field in playbook_run ?

yashmehrotra: returning evaluated val in err msg now

moshloop: I think we should do explict isTruthy and isFalsey checks and then return "Playbook validation failed" error (Without the playbook id which should be known by the calle), or the string response.

playbook/playbook.go
```diff
@@ -65,6 +70,11 @@ func validateAndSavePlaybookRun(ctx context.Context, playbook *models.Playbook,
 		run.Parameters[k] = fmt.Sprintf("%v", v)
 	}
 
+	// Check playbook filters
+	if err := checkPlaybookFilter(ctx, spec, templateEnv); err != nil {
+		return nil, fmt.Errorf("filters did not pass for playbook[%s]: %w", playbook.ID, err)
```
moshloop, line 75:
```suggestion
		return nil, err
```

playbook/runner.go
```diff
@@ -168,6 +168,22 @@ func GetActionForAgent(ctx context.Context, agent *models.Agent) (*ActionForAgen
 	return &output, tx.Commit().Error
 }
 
+func checkPlaybookFilter(ctx context.Context, playbookSpec v1.PlaybookSpec, templateEnv actions.TemplateEnv) error {
+	for _, f := range playbookSpec.Filters {
+		val, err := ctx.RunTemplate(gomplate.Template{Expression: f}, templateEnv.AsMap())
+		if err != nil {
+			return fmt.Errorf("error running playbook filter expression[%s]: %w", f, err)
```
moshloop, line 175:
```suggestion
			return fmt.Errorf("Invalid playbook filter [%s]: %s", f, err)
```

playbook/controllers.go
```diff
@@ -200,6 +196,10 @@ func HandleGetPlaybookParams(c echo.Context) error {
 		return c.JSON(http.StatusInternalServerError, dutyAPI.HTTPError{Error: err.Error(), Message: "failed to unmarshal playbook spec"})
 	}
 
+	if err := checkPlaybookFilter(ctx, spec, env); err != nil {
+		return c.JSON(http.StatusInternalServerError, dutyAPI.HTTPError{Error: err.Error(), Message: "playbook filters did not pass"})
```
moshloop, line 200:
```suggestion
		return c.JSON(http.StatusInternalServerError, dutyAPI.HTTPError{Error: "Playbook validation failed", Message: err.Error()})
```

Issue #860, opened 2024-02-29
fix: pull playbook action job


adityathebe:
`upstream server returned error status[404]: {"error":"not found","message":"agent not found"}`

PR #862, opened 2024-03-01
feat: use the new agent auth middleware & remove hash based reconciliation

adityathebe:
merge after: https://github.com/flanksource/duty/pull/560

resolves: https://github.com/flanksource/mission-control/issues/861
resolves: https://github.com/flanksource/mission-control/issues/860


auth/mock.go
```diff
@@ -28,13 +26,6 @@ func MockMiddleware(next echo.HandlerFunc) echo.HandlerFunc {
 
 		ctx = ctx.WithUser(&models.Person{ID: person.ID, Email: person.Email})
 
-		var agent models.Agent
-		if err := ctx.DB().Where("person_id = ?", person.ID.String()).First(&agent).Error; err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
-			return err
-		} else {
-			ctx = ctx.WithAgent(agent)
-		}
-
```
adityathebe, line 37: In this mock auth middleware, removed setting the agent in the context so we explicitly test that the new agent auth middleware sets it.

The test passes even if this isn't removed.

upstream/upstream_test.go
```diff
@@ -39,46 +27,15 @@ var _ = ginkgo.Describe("Upstream Push", ginkgo.Ordered, func() {
...
-		verifyPushQueue(events.ToPostQEvents(), pushAgent.dataset)
-	})
-
-	ginkgo.It("should track updates & deletes on the event_queue table", func() {
-		start := time.Now()
-
-		modifiedNewDummy := dummy.Logistics
-		modifiedNewDummy.ID = uuid.New()
+		DefaultContext.ClearCache()
+		context.SetLocalProperty("upstream.reconcile.pre-check", "false")
```
moshloop, line 31: We seem to have removed any verification and just assert no error ?

adityathebe: The verification is still there

Issue #857, opened 2024-02-29
Playbooks steps should not run after a step failure


moshloop:
unless `if: always()`

PR #865, opened 2024-03-02
chore: playbooks steps should not run after a step failure

yashmehrotra:
Fixes: #857

### DIFF
--- a/playbook/runner.go
+++ b/playbook/runner.go
@@ -49,20 +49,22 @@ func getNextActionToRun(ctx context.Context, playbook models.Playbook, runID uui
 		}
 	}
 
+	if lastRanAction.Name == "" { // Nothing has run yet
+		return &playbookSpec.Actions[0], nil
+	}
+
 	for i, action := range playbookSpec.Actions {
-		if lastRanAction.Name == "" { // Nothing has run yet
-			return &playbookSpec.Actions[i], nil
-		}
-
-		if action.Name != lastRanAction.Name {
-			continue
-		}
-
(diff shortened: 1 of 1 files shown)


adityathebe:
@moshloop Continuing playbook run on failure was intentional to support for the playbook filters. 

Example: Running a cleanup action or notify action after a previous action failed.

moshloop:
@yashmehrotra - They should only run if we have an explicit filter e.g. `if: always()`

yashmehrotra:
> They should only run if we have an explicit filter e.g. if: always()
@moshloop this is handled in the change

adityathebe:
`if: always()` isn't the only case where we need to run after failure. eg: we also have `failure()`. And it's possible that the current action doesn't have those filters but maybe other actions down the line may have it.

If users want to **not** run actions after failure then they can use the `if: success()` filter. So I don't think we need to make any change in what we currently have.

If `if: success()` is a hassel, then we could add a new field to the playbook that's equivalent to `set -e` in shells.

moshloop:
@adityathebe - I think we should default to the way Github does it, we can skip steps until we find one that will run - we don't need to break on failure

yashmehrotra:
@moshloop @adityathebe 

For skipping playbook actions on failure, there can be cases when users have a filter and want to mix that with always(), like `filter: always() && config.class == "Pod"` so instead of checking and evaluating all, I am just returning the next action with filter and letting the filter function do the deciding

if the playbook has no action filters it will stop at failure

Issue #859, opened 2024-02-29
Add properties endpoint


moshloop:
That uses ctx.Properties() )=(merging db and local prioerties)

PR #907, opened 2024-03-19
feat: properties endpoint

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/859

### DIFF
--- a/echo/serve.go
+++ b/echo/serve.go
@@ -8,6 +8,8 @@ import (
 
 	"github.com/flanksource/commons/logger"
 	cutils "github.com/flanksource/commons/utils"
+	"github.com/flanksource/duty/api"
+	"github.com/flanksource/commons/collections"
 	"github.com/flanksource/duty/context"
 	"github.com/flanksource/duty/schema/openapi"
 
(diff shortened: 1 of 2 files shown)


echo/serve.go
```diff
@@ -160,3 +165,19 @@ func ModifyKratosRequestHeaders(next echov4.HandlerFunc) echov4.HandlerFunc {
 		return next(c)
 	}
 }
+
+func Properties(c echov4.Context) error {
+	ctx := c.Request().Context().(context.Context)
+
+	props, err := db.GetProperties(ctx)
```
moshloop, line 172: Can we make this return a list 

```
properties:
   - name: 
      value: 
      description:
      type: (string|bool|duration|int) 
      source: (local|db)
```

For now the description and type can be empty and string

adityathebe:
```
[
  {
    "description": "",
    "name": "check.http.disabled",
    "source": "db",
    "type": "",
    "value": "true"
  },
  {
    "description": "",
    "name": "check.kubernetes.disabled",
    "source": "local",
    "type": "",
    "value": "true"
  }
]

```

PR #930, opened 2024-03-26
fix: support custom notification(connection) without a name

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/918


api/notifications.go
```diff
@@ -15,9 +15,8 @@ import (
 // the system's SMTP credentials.
 const SystemSMTP = "smtp://system/"
 
-// +kubebuilder:object:generate=true
 type NotificationConfig struct {
-	Name       string            `json:"name"`                                 // A unique name to identify this notification configuration.
+	Name       string            `json:"name,omitempty"`                       // A unique name to identify this notification configuration.
```
adityathebe, line 19: @moshloop Still need the name field because the team spec supports multiple custom notifications. Let me know if that should be removed as well.

PR #945, opened 2024-04-03
feat: make playbook & run objects accessible in templates

adityathebe:
Required by the "Request namespace access" playbook to be used in the rolebinding name.

### DIFF
--- a/playbook/actions/actions.go
+++ b/playbook/actions/actions.go
@@ -4,12 +4,13 @@ import "github.com/flanksource/duty/models"
 
 // TemplateEnv defines the config and component passed to a playbook run action.
 type TemplateEnv struct {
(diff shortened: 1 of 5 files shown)


playbook/actions/actions.go
```diff
@@ -4,24 +4,26 @@ import "github.com/flanksource/duty/models"
...
-		"config":    nil,
-		"user":      nil,
-		"env":       t.Env,
-		"params":    t.Params,
+		"check":       nil,
+		"component":   nil,
+		"config":      nil,
+		"user":        nil,
+		"env":         t.Env,
+		"playbook_id": t.PlaybookID,
```
moshloop, line 25: I think we need the playbook and playbook_run objects

Issue #960, opened 2024-04-19
Implement cleanup for agent deletion


yashmehrotra:
When we delete an agent via UI, we get asked if we want to cleanup everything associated with the agent. Clicking on it does not do anything as that flow hasn't been implemented

cURL:
```
curl 'https://app.flanksource.com/api/db/agents?id=eq.018ef600-0cb0-9a27-d293-8399b9705fbd'
-X PATCH 
--data-raw '{"deleted_at":"now()","cleanup":true}'
```

Need to implement a middleware which intercepts this, and if cleanup is true, mark configs, components, checks etc. with that agent_id as deleted

yashmehrotra:
Fixed via https://github.com/flanksource/duty/pull/696

PR #961, opened 2024-04-19
feat: agent resource cleanup after deletion

yashmehrotra:
Fixes: #960

### DIFF
--- a/db/middleware.go
+++ b/db/middleware.go
@@ -2,6 +2,7 @@ package db
 
 import (
 	"fmt"
+	"net/http"
 	"net/url"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


db/middleware.go
```diff
@@ -87,6 +91,43 @@ func SearchQueryTransformMiddleware() func(echo.HandlerFunc) echo.HandlerFunc {
...
+			if !reqBody.Cleanup || reqBody.DeletedAt != "" {
+				return next(c)
+			}
+
+			if _, err := uuid.Parse(agentID); err != nil {
+				return api.WriteError(c, api.Errorf(api.EINVALID, "unable to parse agent_id (%s) as uuid: %v", agentID, err))
+			}
+
+			ctx := c.Request().Context().(context.Context)
+			if err := cleanupAgentResources(ctx, agentID); err != nil {
```
adityathebe, line 123: @yashmehrotra couldn't we make this a postgresql func & frontend simply calls that via postgREST?

Also not sure why we're not using a regular handler and going with a middleware.

yashmehrotra: > Also not sure why we're not using a regular handler and going with a middleware.
We have to "intercept" db calls via postgrest so that their normal functionality retains intact

> couldn't we make this a postgresql func & frontend simply calls that via postgREST?

I think thats a better idea, a func would mark everything deleted and it all would happen in one transaction

@moshloop What do you think ?

moshloop: Cleanup agent can be a db function, but I think there should be a trigger on on agents that fires a agent.cleanup async event that calls that function or runs the queries. 

The UI can also call that function, but we shouldn't rely on it

PR #999, opened 2024-05-03
feat: cleanup stale running job histories

adityathebe:


jobs/job_history_cleanup.go
```diff
@@ -1,11 +1,34 @@
...
+	"github.com/flanksource/duty/models"
 )
 
-func CleanupJobHistoryTable(ctx job.JobRuntime) error {
-	res := ctx.DB().Exec("DELETE FROM job_history where now() - created_at > interval '30 day'")
-	ctx.History.SuccessCount = int(res.RowsAffected)
-	return res.Error
+var cleanupStaleJobHistory = &job.Job{
+	Name:       "CleanupStaleJobHistory",
+	Schedule:   "@every 4h",
```
moshloop, line 13:
```suggestion
	Schedule:   "@every 24h",
```

jobs/job_history_cleanup.go
```diff
@@ -1,11 +1,34 @@
...
+	Retention:  job.RetentionFew,
+	RunNow:     true,
+	Fn: func(ctx job.JobRuntime) error {
+		staleHistoryMaxAge := ctx.Properties().Duration("job.history.maxAge", time.Hour*24*30)
+		count, err := job.CleanupStaleHistory(ctx.Context, staleHistoryMaxAge, "", "")
+		if err != nil {
+			return fmt.Errorf("error cleaning stale job histories: %w", err)
+		}
+
+		staleRunningHistoryMaxAge := ctx.Properties().Duration("job.history.running.maxAge", time.Hour*4)
```
moshloop, line 25: I don't think we want to delete them, rather mark them as stale, also not sure how this would handle very long running jobs

adityathebe: > not sure how this would handle very long running jobs

I think our best option is to take an educated guess for the "mark as stale" job interval. I think 4 hours is fine as I don't expect any of our job to run for that long.

adityathebe: Also, if we aren't actually deleting those histories and just marking them stale **while the job is actually running**, then I think the job will eventually correct the history when it completes.

PR #1003, opened 2024-05-08
fix: success reporting in event queue cleanup job

adityathebe:
remove stuff related to `push_queue.create`


jobs/event_queue.go
```diff
@@ -3,44 +3,20 @@ package jobs
 import (
 	"time"
 
-	"github.com/flanksource/commons/logger"
 	"github.com/flanksource/duty/job"
 )
 
+const eventQueueStaleAge = time.Hour * 24 * 30
```
moshloop, line 9: Can we make this a property

Issue #1001, opened 2024-05-06
Notification circuit breakers


moshloop:
`notifications.max.window=4h and `notifications.max.count=50`

PR #1011, opened 2024-05-15
feat: sliding window rate limiter on notifications

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/1001

### DIFF
--- a/notification/events.go
+++ b/notification/events.go
@@ -4,7 +4,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"sync"
+	"time"
 
+	sw "github.com/RussellLuo/slidingwindow"
+	"github.com/flanksource/commons/logger"
 	"github.com/flanksource/duty/context"
 	"github.com/flanksource/duty/models"
 	"github.com/flanksource/duty/query"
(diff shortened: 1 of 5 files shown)


notification/events.go
```diff
@@ -28,6 +41,38 @@ func RegisterEvents(ctx context.Context) {
 	events.RegisterAsyncHandler(sendNotifications, 1, 5, api.EventNotificationSend)
 }
 
+func getOrCreateRateLimiter(ctx context.Context, notificationID string) (*sw.Limiter, error) {
+	rateLimitersLock.Lock()
```
moshloop, line 45: Can we use a sync.Map instead - we don't want to  keep a lock while perform db operations.

notification/events.go
```diff
@@ -52,6 +97,17 @@ func addNotificationEvent(ctx context.Context, event postq.Event) error {
...
 		}
 
+		rateLimiter, err := getOrCreateRateLimiter(ctx, id)
+		if err != nil {
+			return fmt.Errorf("failed to create rate limiter: %w", err)
+		}
+
+		if !rateLimiter.Allow() {
+			// rate limited notifications are simply dropped.
+			logger.Debugf("notification(%s) rate limited for event=%s", id, event.Name)
```
moshloop, line 107:
```suggestion
			logger.Warnf("notification(%s) rate limited for event=%s", id, event.Name)
```

notification/events.go
adityathebe: @moshloop we want to lock the db operation in here though. For a sudden burst of event, we only want to make one db query and create one rate limiter for a notification.

This only happens once throughout the lifetime of the process as we'll then hit the cache.

PR #1021, opened 2024-05-20
fix: don't run push_queue.delete consumer on upstream servers

adityathebe:


cmd/server.go
```diff
@@ -31,12 +31,10 @@ import (
 	"github.com/flanksource/incident-commander/notification"
 	"github.com/flanksource/incident-commander/rbac"
 	"github.com/flanksource/incident-commander/teams"
+	"github.com/flanksource/incident-commander/upstream"
 
 	// register event handlers
-	_ "github.com/flanksource/incident-commander/incidents/responder"
```
adityathebe, line 36: These are already imported. Was getting warning for duplicate imports.

cmd/server.go
```diff
@@ -127,6 +125,10 @@ var Serve = &cobra.Command{
 
 		go jobs.Start(ctx)
 
+		if api.UpstreamConf.Valid() {
+			upstream.RegisterEvents(ctx)
```
moshloop, line 129: I think if the upstreamconf is empty we should have a noop handler

Issue #993, opened 2024-05-02
handle cached checks during notification templates


adityathebe:
https://github.com/flanksource/mission-control/blob/bb56a44a8affce5cf5f305f7a57d2d45711215b3/notification/events.go#L131

When templating notifications, we pass in the check from the cache.
The notification template `Check {{.check.name}} is {{.check.status}}` therefore has the same status whether the check passes or fails.

PR #1027, opened 2024-05-21
fix: stale statuses for checks & stale health fro configs & components on notification templates

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/993

### DIFF
--- a/notification/events.go
+++ b/notification/events.go
@@ -12,6 +12,7 @@ import (
 	"github.com/flanksource/duty/context"
 	"github.com/flanksource/duty/models"
 	"github.com/flanksource/duty/query"
+	"github.com/flanksource/duty/types"
 	"github.com/flanksource/incident-commander/api"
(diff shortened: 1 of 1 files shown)


notification/events.go
```diff
@@ -231,6 +232,10 @@ func getEnvForEvent(ctx context.Context, event postq.Event, properties map[strin
 			return nil, fmt.Errorf("failed to get check status for check(%s/%s): %w", checkID, lastRuntime, err)
 		}
 
+		// The check that we supply on the template needs to have the status set
+		// to the status corresponding to the time of event.
+		check.Status = models.CheckHealthStatus(lo.Ternary(checkStatus.Status, models.CheckStatusHealthy, models.CheckStatusUnhealthy))
```
moshloop, line 237: Don't we need to cater for health,status,reason/description etc..

PR #1029, opened 2024-05-22
fix(notification): apply the rate limit only after filter passes

adityathebe:


notification/events.go
```diff
@@ -129,7 +118,18 @@ func addNotificationEvent(ctx context.Context, event postq.Event) error {
...
 
+		rateLimiter, err := getOrCreateRateLimiter(ctx, id)
+		if err != nil {
+			return fmt.Errorf("failed to create rate limiter: %w", err)
+		}
+
 		for _, payload := range payloads {
+			if !rateLimiter.Allow() {
+				// rate limited notifications are simply dropped.
+				logger.Warnf("notification(%s) rate limited for event=%s", id, event.Name)
```
moshloop, line 129: We probably should have a prometheus counter for this.

notification/events.go
```diff
@@ -129,7 +118,19 @@ func addNotificationEvent(ctx context.Context, event postq.Event) error {
...
+				rateLimitedCounter.WithLabelValues(id).Inc()
```
moshloop, line 130: We can just use ctx.Counter("notifications_rate_limited").Incr()

PR #1047, opened 2024-05-28
feat: add gitlab to connection CRD

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/1040


db/connections.go
```diff
@@ -98,6 +98,11 @@ func PersistConnectionFromCRD(ctx context.Context, obj *v1.Connection) error {
 		dbObj.Password = obj.Spec.GitHub.PersonalAccessToken.String()
 	}
 
+	if obj.Spec.GitLab != nil {
+		dbObj.Type = models.ConnectionTypeGitlab
```
moshloop, line 102: I think we need an optional URL here as well - Github and Azure devops should also have

adityathebe: @moshloop why do we need urls for GitHub connection?

PR #1048, opened 2024-05-28
feat: accept POST data into playbook webhook

yashmehrotra:


playbook/controllers.go
```diff
@@ -25,12 +26,26 @@ type RunResponse struct {
 	StartsAt string `json:"starts_at"`
 }
 
+type WebhookRequest struct {
+	Data types.JSONMap `json:"data"`
```
moshloop, line 30: I think we need headers, params (query Params), and content - Also rename date to json - which should be raw json as you could push an array

PR #1057, opened 2024-05-31
feat: auto inject gitops env vars when the playbook has gitops action

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/1053

### DIFF
--- a/playbook/template.go
+++ b/playbook/template.go
@@ -1,0 +1,110 @@
+package playbook
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/flanksource/duty/context"
+	"github.com/flanksource/duty/models"
+	"github.com/flanksource/duty/query"
+	v1 "github.com/flanksource/incident-commander/api/v1"
+)
+
+type GitOpsEnvKustomize struct {
+	Path string `json:"path"`
+}
+
+func (t *GitOpsEnvKustomize) AsMap() map[string]any {
+	return map[string]any{
+		"path": t.Path,
+	}
+}
+
+type GitOpsEnvGit struct {
+	File   string `json:"file"`
+	URL    string `json:"url"`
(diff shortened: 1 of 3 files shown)


playbook/template.go
```diff
@@ -0,0 +1,110 @@
...
+		return nil, err
+	}
+	_ = ci
+
+	var gitOpsEnv GitOpsEnv
+
+	gitRepos := query.TraverseConfig(ctx, run.ConfigID.String(), "Kubernetes::Kustomization/Kubernetes::GitRepository", string(models.RelatedConfigTypeIncoming))
+	if len(gitRepos) > 0 {
+		var config map[string]any
+		if err := json.Unmarshal([]byte(*gitRepos[0].Config), &config); err != nil {
```
moshloop, line 77: Can we use https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured#NestedStringNoCopy, - I think it will look better

playbook/template.go
```diff
@@ -0,0 +1,110 @@
...
+		var config map[string]any
+		if err := json.Unmarshal([]byte(*kustomization[0].Config), &config); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal config: %w", err)
+		}
+		gitOpsEnv.Kustomize.Path = config["spec"].(map[string]any)["path"].(string)
+
+		if spec, ok := config["spec"].(map[string]any); ok {
+			gitOpsEnv.Kustomize.Path = spec["path"].(string)
+		}
+	}
```
moshloop, line 107:
```suggestion
		gitOpsEnv.Kustomize.Path = query.TraverseConfig(ctx, run.ConfigID.String(), "Kubernetes::Kustomization", string(models.RelatedConfigTypeIncoming), "spec", "path")
```

PR #1091, opened 2024-06-14
fix: use POST for /catalog/changes

adityathebe:
https://github.com/flanksource/flanksource-ui/pull/2012#issuecomment-2167654161


moshloop:
@adityathebe can we add an additional POST method to maintain backward compatibility

PR #1146, opened 2024-07-08
Rbacv2

moshloop:


echo/mock.go
```diff
@@ -11,7 +11,7 @@ import (
 // mockAuthMiddleware doesn't actually authenticate since we never store auth data.
 // It simply ensures that the requested user exists in the DB and then attaches the
 // users's ID to the context.
-func MockMiddleware(next echo.HandlerFunc) echo.HandlerFunc {
```
yashmehrotra, line 14: I think this should remain in auth package, echo pkg shouldn't be interacting with db and the logic should be in `Auth` package along with other auth middlewares

rbac/init.go
```diff
@@ -11,56 +12,152 @@ import (
...
+	ObjectCatalog          = "catalog"
+	ObjectConnection       = "connection"
+	ObjectDatabase         = "database"
+	ObjectDatabaseIdentity = "database.identities"
+	ObjectAuthConfidential = "database.kratos"
+	ObjectDatabasePublic   = "database.public"
+	ObjectDatabaseSettings = "database.config_scrapers"
+	ObjectDatabaseSystem   = "database.system"
+	ObjectIncident         = "incident"
+	ObjectMonitor          = "database.monitor"
```
yashmehrotra, line 152: The `database.` prefix is to map directly to tables, we don't have `monitor` and `system` tables

echo/mock.go
moshloop: This is due to an import cycle

Issue #673, opened 2023-12-07
[Saas] Add ability to download kubeconfig to talk to vcluster 


moshloop:
As part of the tenant chart:
- Create a new service account with permissions only to the CRD's, Configmaps and Secrets 
- Create a service-account secret: https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#create-token

- Add an api endpoint `/api/kubeconfig` to retrieve the kubeconfig - generate it and fill in the token from the secret.
- Add an api endpoint  `/api/kube-proxy` that forward traffic to `kubernetes.default.svc` stripping off the path

yashmehrotra:
Approach:
Support Whatever is supported by flux kustomize
Start with basic git://
Strategic merge patches to delete/whitelist

Gitops data in a mission-control table
Once added:
    - Commit back to saas-prod, use playbooks repo
    - Heart beat chart will create cm in workload mc instance
    - Create virtual scraper, insert CI
    - Add CI based events on playbook (on catalog.create update, run these, on delete remove files) see gitops playbook action template
    - Dont create obj directly, create config items
    - Using gitops playbook actions, push it saas prod

PR #1163, opened 2024-07-15
feat: kubeconfig download api endpoint

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/673


echo/kube_config_download.go
```diff
@@ -0,0 +1,76 @@
...
+
+func KubeProxyTokenMiddleware(next echo.HandlerFunc) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		path := "/var/run/secrets/kubernetes.io/serviceaccount/token"
+		saToken, err := os.ReadFile(path)
+		if err != nil {
+			return fmt.Errorf("failed to read service account token: %w", err)
+		}
+
+		c.Request().Header.Set("Authorization", fmt.Sprintf("Bearer %s", saToken))
```
moshloop, line 73: Also need to add the impersonation headers  - they shouldn't have access to mission-control-sa which allows reading secrets / creating pods etc..

PR #1212, opened 2024-07-31
feat: create access token for kubeconfig

adityathebe:


echo/kube_config_download.go
```diff
@@ -32,27 +35,41 @@ current-context: {{.ContextName}}
...
 		return fmt.Errorf("failed to parse public web url")
 	}
 	parsed.Path = path.Join(parsed.Path, "kubeproxy")
 
+	seed, err := rand.GenerateRandHex(32)
+	if err != nil {
+		return err
+	}
+
+	token, err := db.CreateAccessToken(ctx, ctx.User().ID, ctx.User().Email, seed, time.Hour*24*90)
```
moshloop, line 63: I think this should have a nil expiry

echo/kube_config_download.go
```diff
@@ -32,27 +35,41 @@ current-context: {{.ContextName}}
...
+	if err != nil {
+		return fmt.Errorf("failed to create a new access token: %w", err)
+	}
+
 	kcd := kubeConfigData{
 		Server:      parsed.String(),
-		ClusterName: "kubernetes",
+		ClusterName: "tenant",
```
moshloop, line 70: Does this get replaced with mission-control-{tenant-slug}

echo/kube_config_download.go
```diff
@@ -32,27 +34,43 @@ current-context: {{.ContextName}}
...
+	if err != nil {
+		return fmt.Errorf("failed to create a new access token: %w", err)
+	}
+
+	// TODO: use the actual tenant slug once we pass it on from the chart
+	tenantSlug := strings.TrimSuffix(parsed.Hostname(), ".flanksource.com")
 	kcd := kubeConfigData{
 		Server:      parsed.String(),
-		ClusterName: "kubernetes",
+		ClusterName: fmt.Sprintf("mission-control-tenant-%s", tenantSlug),
```
moshloop, line 71:
```suggestion
		ClusterName: fmt.Sprintf("mission-control-%s", tenantSlug),
```

Issue #1297, opened 2024-08-20
Notifications - Repeat interval


moshloop:
Do not send notifications if they have been fired within interval
```
- groupBy:
      - notification_id
      - source_event
      - resource_id
   repeatInterval: 1d
```      

Increment count on notification_send_history instead

PR #1314, opened 2024-08-25
feat: add repeat interval check for notifications

yashmehrotra:
Blocked on: https://github.com/flanksource/mission-control/pull/1312

Fixes: https://github.com/flanksource/mission-control/issues/1297

### DIFF
--- a/notification/events.go
+++ b/notification/events.go
@@ -96,25 +96,25 @@ func checkRepeatInterval(ctx context.Context, n NotificationWithSpec, event mode
 			clauses = append(clauses, clause.Eq{Column: g, Value: val})
 		}
 	}
-	var lastSent time.Time
 	if len(clauses) > 0 {
-		tx := ctx.DB().Model(&models.NotificationSendHistory{}).Clauses(clauses...).Select("created_at").Scan(&lastSent)
+		interval, err := text.ParseDuration(n.RepeatInterval)
+		if err != nil {
+			return false, fmt.Errorf("error parsing repeat interval[%s] to time.Duration: %w", n.RepeatInterval, err)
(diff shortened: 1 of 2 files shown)


notification/events.go
```diff
@@ -111,11 +152,24 @@ func (t *notificationHandler) addNotificationEvent(ctx context.Context, event mo
 			continue
 		}
 
+		// TODO: How does this get unset ?
```
yashmehrotra, line 155: @moshloop @adityathebe We add this error to a notification but we where do we unset it, because it gets skipped when it is set

adityathebe: there's a postgres trigger

```
reset_notification_error_before_update_trigger BEFORE UPDATE ON notifications FOR EACH ROW EXECUTE FUNCTION reset_notification_error_before_update()
```

moshloop: We should probably just update the comment here with that detail

notification/events.go
```diff
@@ -81,6 +83,45 @@ type notificationHandler struct {
...
+		}
+		if lastSent.IsZero() {
+			return false, fmt.Errorf("last sent not found for notification[%s]", n.ID)
+		}
+
+		interval, err := text.ParseDuration(n.RepeatInterval)
+		if err != nil {
+			return false, fmt.Errorf("error parsing repeat interval[%s] to time.Duration: %w", n.RepeatInterval, err)
+		}
+		if time.Since(lastSent) <= *interval {
```
moshloop, line 117: We are comparing db time and local time here - We should move this to the query

yashmehrotra: Updated

Issue #923, opened 2024-03-22
Slack Notifications - Support Blocks


moshloop:
https://api.slack.com/block-kit

Create some go text template helpers:

- ` slackFields (dict "Type" "Computer (laptop)", "When" , "Submitted Aut 10" | slackSection`
- `{{ if eq type "slack"}}`
- Update existing default templates to use a slack formatted message 

```json
{
	"blocks": [
		{
			"type": "section",
			"text": {
				"type": "mrkdwn",
				"text": ":red_circle: *default/pod-info* is _unhealthy_"
			}
		},
		{
			"type": "section",
			"text": {
				"type": "mrkdwn",
				"text": ">WARNING failed to watch kubernetes resources: failed to apply kube config: could not find secret /dev-kubeconfig: an empty namespace may not be set when a resource name is provided"
			}
		},
		{
			"type": "section",
			"fields": [
				{
					"type": "mrkdwn",
					"text": "*cluster*\nkubernetes"
				},
				{
					"type": "mrkdwn",
					"text": "*namespace*\ncrossplane-system",
					"verbatim": true
				},
				{
					"type": "mrkdwn",
					"text": "*apiVersion*\nsource.toolkit.fluxcd.io/v1",
					"verbatim": true
				},
				{
					"type": "mrkdwn",
					"text": "*kustomize.toolkit.fluxcd.io/name*\ncommon",
					"verbatim": true
				},
				{
					"type": "mrkdwn",
					"text": "*kustomize.toolkit.fluxcd.io/namespace*\nflux-system",
					"verbatim": true
				},
				{
					"type": "mrkdwn",
					"text": "*kustomize.toolkit.fluxcd.io/namespace*\nflux-system",
					"verbatim": true
				}
			]
		},
		{
			"type": "actions",
			"elements": [
				{
					"type": "button",
					"text": {
						"type": "plain_text",
						"text": "Config",
						"emoji": true
					},
					"value": "click_me_123",
					"action_id": "actionId-0"
				}
			]
		}
	]
}
```

PR #1325, opened 2024-08-28
feat: use new Slack module to send slack notifications

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/923

- needed quite a bit of refactoring because the templating of the notification title & body needs to be pushed to the very end of the pipeline instead of the very beginning. This is because now we need to template the message only after we've figured out the shoutrrr service we're sending to.

### DIFF
--- a/notification/send.go
+++ b/notification/send.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/flanksource/commons/collections"
 	"github.com/flanksource/commons/utils"
 	"github.com/flanksource/duty/context"
 	"github.com/flanksource/duty/models"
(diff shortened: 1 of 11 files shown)


notification/send.go
```diff
@@ -114,17 +101,78 @@ func SendNotification(ctx *Context, payload NotificationEventPayload, celEnv map
...
+		}
+
+		shoutrrrURL = connection.URL
+		data.Properties = collections.MergeMap(connection.Properties, data.Properties)
+	}
+
+	if connection != nil && connection.Type == models.ConnectionTypeSlack {
+		// We know we are sending to slack.
+		// Send the notification with slack-api and don't go through Shoutrrr.
+		celEnv["outgoing_channel"] = "slack"
```
moshloop, line 160:
```suggestion
		celEnv["channel"] = "slack"
```

notification/templates/check.failed
```diff
@@ -0,0 +1,40 @@
+{{ if eq outgoing_channel "slack"}}
+{
+	"blocks": [
+		{{slackSectionTextMD (printf `:red_circle: *%s* is _unhealthy_` .check.name)}},
+		{{ if .status.message}}{{slackSectionTextMD status.message}},{{end}}
+		{
+			"type": "section",
+			"fields": [
+				{{slackSectionTextFieldMD (printf `*Canary*: %s` .canary.name) }},
+				{{slackSectionTextFieldMD (printf `*Namespace*: %s` .canary.namespace) }}
```
moshloop, line 10: We need the canary/checks labels as well

adityathebe:
image

PR #1338, opened 2024-09-02
fix: save notification repeat interval & groupBy via CRD

adityathebe:


notification/events.go
```diff
@@ -84,40 +84,42 @@ type notificationHandler struct {
...
 	validKeys := map[string]string{
-		"notification_id": n.ID.String(),
-		"resource_id":     event.Properties["id"],
-		"source_event":    event.Name,
+		"resource_id":  event.Properties["id"],
+		"source_event": event.Name,
 	}
-	var clauses []clause.Expression
+
+	clauses := []clause.Expression{clause.Eq{Column: "notification_id", Value: n.ID.String()}}
```
adityathebe, line 94: @moshloop I think we always add this clause since this is a per notification spec.

adityathebe: If we add `count` in here this because an advanced version of our existing rate limitation.

And the existing rate limitation can instead be a global rate limit (or per channel)

moshloop: Repeat Interval is a well understood term - rate limiting serves a different purpose - If a notification doesn't have a repeat interval, or is misconfigured then it will spam channels and therefore is not something that is opt-in, it is there to protect the slack channels / email accounts from spam and endless loops

api/v1/notification_types.go
```diff
@@ -45,6 +45,13 @@ type NotificationSpec struct {
 	// Cel-expression used to decide whether this notification client should send the notification
 	Filter string `json:"filter,omitempty" yaml:"filter,omitempty"`
 
+	// RepeatInterval is the waiting time to resend a notification after it has been succefully sent.
+	RepeatInterval string `json:"rateLimit,omitempty" yaml:"rateLimit,omitempty"`
```
moshloop, line 49:
```suggestion
	RepeatInterval string `json:"repeatInterval,omitempty" yaml:"repeatInterval,omitempty"`
```

api/v1/notification_types.go
```diff
@@ -45,6 +45,13 @@ type NotificationSpec struct {
 	// Cel-expression used to decide whether this notification client should send the notification
 	Filter string `json:"filter,omitempty" yaml:"filter,omitempty"`
 
+	// RepeatInterval is the waiting time to resend a notification after it has been succefully sent.
+	RepeatInterval string `json:"repeateInterval,omitempty" yaml:"repeateInterval,omitempty"`
```
moshloop, line 49:
```suggestion
	RepeatInterval string `json:"repeatInterval,omitempty" yaml:"repeatInterval,omitempty"`
```

notification/events.go
```diff
@@ -84,43 +84,36 @@ type notificationHandler struct {
...
 	for _, g := range n.GroupBy {
 		if val, exists := validKeys[g]; exists {
 			clauses = append(clauses, clause.Eq{Column: g, Value: val})
+		} else {
```
moshloop, line 98: I think we should just group by resource-id, event and notification id by default and handling custom grouping later

Issue #1329, opened 2024-08-29
Slack notification: Handle large number of labels


adityathebe:
```
invalid_blocks
[ERROR] no more than 10 items allowed [json-pointer:/blocks/3/fields]
```

```json
{
  "blocks": [
    {
      "type": "section",
      "text": {
        "type": "mrkdwn",
        "text": ":information_source: *nginx-ingress-ingress-nginx-controller* was _created_"
      }
    },
    {
      "type": "divider"
    },
    {
      "type": "section",
      "fields": [
        {
          "type": "mrkdwn",
          "text": "*Type*: Kubernetes::ConfigMap"
        },
        {
          "type": "mrkdwn",
          "text": "*Status*: Unknown"
        }
      ]
    },
    {
      "fields": [
        {
          "text": "*apiVersion*: v1",
          "type": "mrkdwn",
          "verbatim": true
        },
        {
          "text": "*app.kubernetes.io/instance*: nginx-ingress",
          "type": "mrkdwn",
          "verbatim": true
        },
        {
          "text": "*app.kubernetes.io/name*: ingress-nginx",
          "type": "mrkdwn",
          "verbatim": true
        },
        {
          "text": "*app.kubernetes.io/part-of*: ingress-nginx",
          "type": "mrkdwn",
          "verbatim": true
        },
        {
          "text": "*app.kubernetes.io/version*: 1.11.2",
          "type": "mrkdwn",
          "verbatim": true
        },
        {
          "text": "*helm.toolkit.fluxcd.io/namespace*: nginx-ingress",
          "type": "mrkdwn",
          "verbatim": true
        },
        {
          "text": "*app.kubernetes.io/component*: controller",
          "type": "mrkdwn",
          "verbatim": true
        },
        {
          "text": "*app.kubernetes.io/managed-by*: Helm",
          "type": "mrkdwn",
          "verbatim": true
        },
        {
          "text": "*cluster*: homelab",
          "type": "mrkdwn",
          "verbatim": true
        },
        {
          "text": "*helm.sh/chart*: ingress-nginx-4.11.2",
          "type": "mrkdwn",
          "verbatim": true
        },
        {
          "text": "*helm.toolkit.fluxcd.io/name*: nginx-ingress",
          "type": "mrkdwn",
          "verbatim": true
        },
        {
          "text": "*namespace*: nginx-ingress",
          "type": "mrkdwn",
          "verbatim": true
        }
      ],
      "text": {
        "text": "*Labels*",
[...]

PR #1344, opened 2024-09-03
fix: slack labels in section limit

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/1329

image

### DIFF
--- a/notification/template.go
+++ b/notification/template.go
@@ -3,6 +3,7 @@ package notification
 import (
 	"encoding/json"
 	"fmt"
+	"slices"
 	"strings"
 
 	"github.com/flanksource/duty/models"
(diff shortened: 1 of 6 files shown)


moshloop:
@adityathebe - Can we make sure that tags are always included, (and also sorted first?)

adityathebe:
> @adityathebe - Can we make sure that tags are always included, (and also sorted first?)

@moshloop Isn't that already the case? All labels include the tags. We don't truncate any labels.

I've sorted the labels chronologically but I don't think we can place the tags before labels.
They're just key-vals. Unless the user calls 

```
slackSectionLabels(config.tags, config.labels)
```

but this wouldn't work well with other objects (checks, components) that don't have both tags/labels.

moshloop:
@adityathebe we should just have a slackSectionLabels(config|component|check)

adityathebe:
@moshloop  Made the change 

image

Issue #1346, opened 2024-09-04
Update user api


moshloop:
- Name
- Email
- Role
- Active/Inactive

moshloop:
+ Update the trigger to include updates to the identities table

PR #1347, opened 2024-09-04
feat: /people/update endpoint

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/1346

merge after: https://github.com/flanksource/duty/pull/1012

### DIFF
--- a/echo/users.go
+++ b/echo/users.go
@@ -1,0 +1,98 @@
+package echo
+
+import (
+	"errors"
+	"net/http"
+
+	"github.com/flanksource/duty/api"
+	"github.com/flanksource/duty/context"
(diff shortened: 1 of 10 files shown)


echo/users.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/samber/lo"
+)
+
+type UpdateUserRequest struct {
+	ID string `json:"id" form:"id"`
+
+	FirstName *string `json:"firstName" form:"firstName"`
+	LastName  *string `json:"lastName" form:"lastName"`
+	Email     *string `json:"email" form:"email"`
+	Role      *string `json:"role" form:"role"`
```
moshloop, line 23: Need to cover this as well

PR #1348, opened 2024-09-05
chore: move examples from docs to fixutres

yashmehrotra:


moshloop:
Can we add a test to see if they can be applied to k8s ?

yashmehrotra:
@moshloop You mean to test if the schema is valid ?

PR #1359, opened 2024-09-06
fix: notification repeat interval test

yashmehrotra:


notification/notification_test.go
```diff
@@ -168,7 +167,11 @@ var _ = ginkgo.Describe("Notifications", ginkgo.Ordered, func() {
 		})
 
 		ginkgo.It("should have sent a notification for a config update", func() {
-			err := DefaultContext.DB().Model(&models.ConfigItem{}).Where("id = ?", config.ID).UpdateColumn("config", `{"color": "blue"}`).Error
+			event := models.Event{
+				Name:       "config.updated",
+				Properties: types.JSONStringMap{"id": config.ID.String()},
+			}
+			err := DefaultContext.DB().Create(&event).Error
```
yashmehrotra, line 174: config.updated is only fired when we get a diff type in config_change table, so creating this manually

Issue #1337, opened 2024-09-02
Notification Silences


moshloop:
1. Table

notification_silences
 config/component/check_id/canary_id
 from-until 

2. Create the notification, but don't send it and mark it as silenced 
3. Recursive option via path query
4. Update slack templates / function with a Silence action block - That hits a new /notifications/silence endpoint

PR #1364, opened 2024-09-09
feat: notification silence

adityathebe:
merge after: https://github.com/flanksource/duty/pull/1030

resolves: https://github.com/flanksource/mission-control/issues/1337


api/v1/notification_silence.go
```diff
@@ -0,0 +1,51 @@
...
+type NotificationSilenceResource struct {
+	ConfigID    *string `json:"config_id,omitempty"`
+	CanaryID    *string `json:"canary_id,omitempty"`
+	ComponentID *string `json:"component_id,omitempty"`
+	CheckID     *string `json:"check_id,omitempty"`
+}
+
+// +kubebuilder:object:generate=true
+type NotificationSilenceSpec struct {
+	NotificationSilenceResource `json:",inline" yaml:",inline"`
```
moshloop, line 16: I don't think putting ID's into a crd makes sense - these should be selectors,

db/notifications.go
```diff
@@ -111,3 +142,55 @@ func NotificationSendSummary(ctx context.Context, id string, window time.Duratio
...
+	query := ctx.DB().Debug().Model(&models.NotificationSilence{})
+
+	// Initialize with a false condition,
+	// if no resources are provided, the query won't return all records
+	orClauses := ctx.DB().Where("1 = 0")
+
+	if resources.ConfigID != nil {
+		orClauses = orClauses.Or("config_id = ?", *resources.ConfigID)
+
+		// recursive stuff
```
moshloop, line 158: What does this query look like when generated ?

adityathebe:
```sql
With path_cte AS (
  SELECT path FROM config_items Where id = $1
)
SELECT COUNT(*)
  FROM notification_silences
  CROSS JOIN path_cte
WHERE (config_id = $1 OR (recursive = true AND path_cte LIKE '%s' || config.id::TEXT || '%s'))
  AND from <= NOW()
  AND until >= NOW()
```

moshloop: Do we have a test for this, because I think the path query is wrong - If there is an existing silence on a config_id (parent), then we also silence with config.path like parent.path + "%'

notification/silence.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/flanksource/commons/duration"
+	"github.com/flanksource/duty/context"
+	"github.com/flanksource/duty/models"
+	"github.com/samber/lo"
+)
+
+type SilenceSaveRequest struct {
+	models.NotificationSilenceResource
+	From        time.Time `json:"from"`
+	Until       time.Time `json:"until"`
```
moshloop, line 16: Can we use date math for these

db/notifications.go
adityathebe: > If there is an existing silence on a config_id (parent), then we also silence with config.path like parent.path + "%'

This would only work for root components no? So if you have silenced on a deployment , that would not reciprocate to the pod because the pod's path is `<namespace>.<deployment>.<pod>`. It would work if you've silenced the namespace however.

We do have a test for this. https://github.com/flanksource/mission-control/pull/1364/files#diff-c18508950d90e5dd1426ae0b6a32c3f68477ce565551303c2835a701ffc16b9eR87-R98

A parent is silenced and consequently its child and grandchild is also silenced.

moshloop: The pods path should start with the namespace/deployment path

Issue #1235, opened 2024-08-05
Improve playbook runner speed


adityathebe:
At the moment, runners poll for playbook actions `@every 1m`. Once it runs it, the result is then pushed `@every 15s`.

Due to this, it takes at few minutes to complete a simple playbook with one action.
With more actions, time to complete increases linearly.

### Proposed solutions

1. Long polling

Runners shouldn't run on a schedule. Rather they should be continuously polling for actions in a never ending for loop.

2. Stateful connection

Somehow maintain a stateful TCP connection between the upstream and the runner. 
Runners "subscribe" to upstream on startup and upstream "publishes" to the connection whenever a new action for that agent is inserted to db.

moshloop:
Implement HTTP/2 long polling for maximum portability + channels to incrementally push playbook action / artifacts /

PR #1374, opened 2024-09-10
feat: longpoll for playbook runners

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/1235


playbook/runner/longpoll.go
```diff
@@ -0,0 +1,30 @@
+package runner
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/flanksource/duty/postq/pg"
+)
+
+var LongpollTimeout = time.Minute
```
moshloop, line 10: This should be a property

playbook/runner/agent.go
```diff
@@ -19,7 +20,32 @@ type ActionForAgent struct {
...
+	if action != nil {
+		return action, err
+	}
+
+	// Go into waiting state
+	select {
+	case <-time.After(LongpollTimeout):
+		return &ActionForAgent{}, nil
+
+	case <-ActionNotifyRouter.RegisterRoutes(agent.ID.String()):
```
moshloop, line 38: This would call RegisterRoutes for every action? Shouldn't their be 1 Register call, and then a GetChannelForAgent - Or rename to GetOrCreateChannelForAgent

adityathebe: `RegisterRoutes` returns the existing channel if it was previously registered.

moshloop: Then it should probably be named GetOrCreateChannel

Issue #1406, opened 2024-09-16
Notification - wait delay


moshloop:
i.e. When a notification is fired, wait for a period to re-evalaute health before sending the notification. 

The notification should go into a status of Pending. 

Apply healthy events.

PR #1407, opened 2024-09-17
feat: notification wait for

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/1406


notification/events.go
```diff
@@ -274,7 +275,27 @@ func sendNotifications(ctx context.Context, events models.Events) models.Events
...
+
+			health, err := celEnv.GetResourceHealth(ctx)
+			if err != nil {
+				e.SetError(err.Error())
+				failedEvents = append(failedEvents, e)
+				notificationContext.WithError(err.Error())
+				continue
+			}
+
+			if health != api.EventToHealth(originalEvent.Name) {
```
moshloop, line 292: Need to cater for a notifiucation on warning/unhealthy and the eveny firing on warning and then changing to the other

notification/events.go
```diff
@@ -231,9 +231,10 @@ func addNotificationEvent(ctx context.Context, id string, celEnv map[string]any,
 			continue
 		}
 
-		newEvent := api.Event{
+		newEvent := models.Event{
 			Name:       api.EventNotificationSend,
 			Properties: payload.AsMap(),
+			Delay:      n.WaitFor,
```
moshloop, line 237: I think we still need to insert a pending notification into notification send.history and then update it on final send

PR #1417, opened 2024-09-18
feat: add more details to postgrest spans

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/829

image

### DIFF
--- a/echo/serve.go
+++ b/echo/serve.go
@@ -33,6 +33,7 @@ import (
 	"github.com/labstack/echo/v4/middleware"
 	prom "github.com/prometheus/client_golang/prometheus"
 	"go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho"
+	"go.opentelemetry.io/otel/attribute"
 )
 
 const (
@@ -130,6 +131,7 @@ func New(ctx context.Context) *echov4.Echo {
 		Forward(ctx, e, "/db", dutyApi.DefaultConfig.Postgrest.URL,
 			rbac.DbMiddleware(),
 			db.SearchQueryTransformMiddleware(),
+			postgrestTraceMiddleware,
 		)
 	}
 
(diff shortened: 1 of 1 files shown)


moshloop:
@adityathebe can we make the span name /db/table_name ?

adityathebe:
> @adityathebe can we make the span name /db/table_name ?

@moshloop I don't think the echo otel middleware allows it

https://github.com/open-telemetry/opentelemetry-go-contrib/blob/a6c39275e680e3316303c2f6210d1cc5fb1752cf/instrumentation/github.com/labstack/echo/otelecho/echo.go#L70-L76

echo/serve.go
```diff
@@ -169,6 +171,21 @@ func New(ctx context.Context) *echov4.Echo {
 	return e
 }
 
+func postgrestTraceMiddleware(next echov4.HandlerFunc) echov4.HandlerFunc {
+	return func(c echov4.Context) error {
+		ctx := c.Request().Context().(context.Context)
+
+		table := strings.TrimPrefix(c.Request().URL.Path, "/db/")
+		ctx.GetSpan().SetAttributes(attribute.String("postgrest.table", table))
```
moshloop, line 179:
```suggestion
		ctx.GetSpan().SetAttributes(attribute.String("db.table", table))
```

echo/serve.go
```diff
@@ -169,6 +171,21 @@ func New(ctx context.Context) *echov4.Echo {
...
+
+		for query, values := range c.Request().URL.Query() {
+			ctx.GetSpan().SetAttributes(attribute.String(fmt.Sprintf("postgrest.query.%s", query), values[0]))
```
moshloop, line 182:
```suggestion
			ctx.GetSpan().SetAttributes(attribute.String(fmt.Sprintf("db.query.%s", query), values[0]))
```

moshloop:
@adityathebe it is not that important for now

PR #1423, opened 2024-09-19
chore: add catalog summary materialized view referesh jobs

yashmehrotra:
Depends on https://github.com/flanksource/duty/pull/1073

### DIFF
--- a/jobs/catalog.go
+++ b/jobs/catalog.go
@@ -1,0 +1,29 @@
+package jobs
+
+import (
+	"github.com/flanksource/duty/job"
+)
+
+var RefreshCatalogAnaylsisChangeCount7dView = &job.Job{
+	Name:       "RefreshCatalogAnaylsisChangeCount7dView",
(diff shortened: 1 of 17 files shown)


jobs/catalog.go
```diff
@@ -0,0 +1,29 @@
...
+	Retention:  job.RetentionFew,
+	Singleton:  true,
+	JobHistory: true,
+	RunNow:     true,
+	Fn: func(ctx job.JobRuntime) error {
+		return job.RefreshConfigItemAnalysisChangeCount7d(ctx.Context)
+	},
+}
+
+var RefreshCatalogAnaylsisChangeCount30dView = &job.Job{
```
yashmehrotra, line 19: @moshloop When some first sets up the instance, the materialized view will be empty, and say they run the scrapers etc. after 15 mins, they would have to wait for an hour for catalog page to work since in that case it would be empty

Issue #1428, opened 2024-09-20
Playbook spec runs on are not used


adityathebe:
```yaml
spec:
  runsOn: ['aws']
  actions:
    - name: first
    - name: second
```

Even though the spec says the playbook should run on aws agent, it runs all the actions on the local agent because the actions do not have `runsOn = aws` on them.

PR #1429, opened 2024-09-20
fix: runsOn handling

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/1428

### DIFF
--- a/playbook/runner/runner.go
+++ b/playbook/runner/runner.go
@@ -142,12 +142,26 @@ func CheckDelay(ctx context.Context, playbook models.Playbook, run models.Playbo
 	return false, nil
 }
 
+func getEligibleAgents(spec v1.PlaybookSpec, action *v1.PlaybookAction, run models.PlaybookRun) []string {
+	if run.AgentID != nil {
+		return []string{run.AgentID.String()}
+	}
+
+	if len(action.RunsOn) != 0 {
+		return action.RunsOn
+	}
+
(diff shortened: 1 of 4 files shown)


playbook/runner/runner.go
```diff
@@ -179,24 +189,34 @@ func ScheduleRun(ctx context.Context, run models.PlaybookRun) error {
 		}, action.Name))
 	}
 
-	if run.AgentID == nil && (len(action.RunsOn) == 0 || lo.Contains(action.RunsOn, Main)) {
+	var playbookSpec v1.PlaybookSpec
```
moshloop, line 192: This doesn't cater for an agent specified on the request

playbook/runner/runner.go
```diff
@@ -142,12 +142,26 @@ func CheckDelay(ctx context.Context, playbook models.Playbook, run models.Playbo
 	return false, nil
 }
 
+func getEligibleAgents(spec v1.PlaybookSpec, action *v1.PlaybookAction, run models.PlaybookRun) []string {
+	if run.Parameters != nil && run.Parameters["agent_id"] != "" {
```
moshloop, line 146: Where does this agent_id come from? RunParams has a AgentID field which maps to run.AgentId = we can't inject it into parameters, as then they are no longer valid (extra param)

adityathebe: ah I misread `RunParams.AgentID` to be a run parameter.

Issue #1169, opened 2024-07-16
ABAC


moshloop:
See https://casbin.org/docs/abac

Where request is:

```yaml
playbook:
  name: name
  id: 
  labels: {}
config:
  name: name
  owners: []
  id
  type: Kubernetes::Pod 
  tags: {}
  path: []
teams: [] # teams the user is member of
```

Which would allow policies to be added like:

```yaml
kind: Permission
metadata: 
   uid: abc
permissions:
   - action: playbook:run
     team: everyone
     inherited: true
     config:
         #...selectorFields
         id: def # id of aws cluster
    - action: playbook:run
       owner: true
    - action: playbook:run
       team: SRE Team
       playbook: name or ID
```

Which corresponds to: 

`p, r.playbook.id == 'abc' && 'def' in r.config.path`
`p, r.playbook.id == 'abc' && r.sub.id in r.config.owners`
`p, r.playbook.id == 'abc' && 'everyone' in  r.sub.teams && ' 'def' in r.config.path`

And then from a UI perspective on a config item there would be a permission tab/table with:

|action|playbook|User/Team|
|---|---| ---|
|playbook:run| *| editors (default global policy) |

And on the playbook permissions tab
|action|playbook|User/Team|
|---|---| ---|
|playbook:run| *| editors (default policy) 
|playbook:run| playbook id| `$owners` |
|playbook:run| playbook id| SRE Team|

The permission table would need:

- playbook_id (optional, if empty all playbooks)
- config_id (optional if empty all configs)
- config_selector (
- inherited (id in r.config.id) or (id in r.config.path)
- deny
- action
- user
- team

PR #1440, opened 2024-09-23
feat: permissions adapter for casbin

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/1169


playbook/approval.go
```diff
@@ -52,6 +52,11 @@ func requiresApproval(spec v1.PlaybookSpec) bool {
 
 func approveRun(ctx context.Context, run *models.PlaybookRun) error {
 	approver := ctx.User()
+	if abacResource, err := runToABACResource(ctx, *run); err != nil {
+		return ctx.Oops().Wrap(err)
+	} else if !rbac.HasPermission(ctx, approver.ID.String(), abacResource, rbac.ActionPlaybookApprove) {
```
moshloop, line 57: i think we should add `GetPermissionObjects() []Object` to  a models/accessors.go 

for a playbook run  initially we just return the playbooks, but in subsequent phases we would return the objects referred to as well - the permission check than applies to all.

PR #1444, opened 2024-09-24
fix: deletion logic in push topology

yashmehrotra:
Depends on: https://github.com/flanksource/duty/pull/1088


yashmehrotra:
@moshloop Merging this

Issue #1446, opened 2024-09-24
Templating error for canary failures 


moshloop:
```yaml
---
error:
  context:
    template: "<p>{{ if eq channel &ldquo;slack&rdquo;}}\n{</p>\n\n<pre><code>&quot;blocks&quot;:
      [\n\t{{slackSectionTextMD (printf `:red_circle: *%s* is _unhealthy_` .check.name)}},\n{&quot;type&quot;:
      &quot;divider&quot;},\n\t{{ if .status.error}}{{slackSectionTextMD status.error}},{{end}}\n\t{\n\t\t&quot;type&quot;:
      &quot;section&quot;,\n\t\t&quot;fields&quot;: [\n\t\t\t{{slackSectionTextFieldMD
      (printf `*Canary*: %s` .canary.name) }},\n\t\t\t{{slackSectionTextFieldMD (printf
      `*Namespace*: %s` .canary.namespace) }}\n\t\t\t{{if ne .agent.name &quot;local&quot;}}\n\t\t\t\t,{{slackSectionTextFieldMD
      (printf `*Agent*: %s` .agent.name) }}\n\t\t\t{{end}}\n\t\t]\n\t},\n\t{{ if .check.labels}}{{slackSectionLabels
      .check}},{{end}}\n\t{{ slackURLAction &quot;View Health Check&quot; .permalink
      &quot;\U0001F515 Silence&quot; .silenceURL}}\n]\n</code></pre>\n\n<p>}\n{{ else
      }}\nCanary: {{.canary.name}}\n{{if .agent}}Agent: {{.agent.name}}{{end}}\nError:
      {{.status.error}}\n{{labelsFormat .check.labels}}</p>\n\n<p><a href=\"{{.permalink}}\"
      target=\"_blank\">Reference</a>\n{{end}}</p>\n"
  error: 'template: :1: unexpected "&" in operand'
  stacktrace: |-
    Oops: template: :1: unexpected "&" in operand
      --- at /go/pkg/mod/github.com/flanksource/gomplate/v3@v3.24.32/template.go:295 goTemplate()
      --- at /go/pkg/mod/github.com/flanksource/gomplate/v3@v3.24.32/structtemplater.go:161 StructTemplater.Template()
      --- at /go/pkg/mod/github.com/flanksource/gomplate/v3@v3.24.32/structtemplater.go:51 StructTemplater.StructField()
[...]

PR #1449, opened 2024-09-24
fix: templating of notifications for SMTP

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/1446

This happend for SMTP notifications.

We need to first template and then convert the markdown to HTML.

Else the `markdown -> HTML` conversion will break all the go templating logic.

### DIFF
--- a/notification/cel.go
+++ b/notification/cel.go
@@ -10,7 +10,8 @@ import (
 )
 
 type celVariables struct {
-	Agent *models.Agent
+	Agent   *models.Agent
+	Channel string
 
 	ConfigItem  *models.ConfigItem
 	Component   *models.Component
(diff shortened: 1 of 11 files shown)


notification/shoutrrr.go
```diff
@@ -64,6 +64,12 @@ func shoutrrrSend(ctx *Context, celEnv map[string]any, shoutrrrURL string, data
 		return "", fmt.Errorf("failed to extract service name: %w", err)
 	}
 
+	celEnv["channel"] = service
```
moshloop, line 67: Can we add a test that fails on the previous implementation ?

notification/shoutrrr_test.go
```diff
@@ -58,3 +60,23 @@ var _ = ginkgo.Describe("Notification properties", func() {
 		})
 	}
 })
+
+var _ = ginkgo.Describe("Shoutrrr", func() {
+	ginkgo.It("should template smtp", func() {
```
moshloop, line 65: Should we not be testing all the default templates ?

adityathebe: Oh this was mainly to test the SMTP markdown -> html conversion.

I can add tests for all the templates.

notification/events.go
```diff
@@ -338,13 +338,13 @@ func isHealthReportable(events []string, previousHealth, currentHealth models.He
 	return healthDegraded
 }
 
-// getEnvForEvent gets the environment variables for the given event
+// GetEnvForEvent gets the environment variables for the given event
 // that'll be passed to the cel expression or to the template renderer as a view.
-func getEnvForEvent(ctx context.Context, event models.Event, properties map[string]string) (*celVariables, error) {
```
adityathebe, line 343: All the properties that we need are in the event itself. The 3rd param isn't required.

adityathebe:
@moshloop added tests for all templates.

notification/cel.go
```diff
@@ -85,6 +86,10 @@ func (t *celVariables) AsMap() map[string]any {
 		"author":  lo.FromPtr(t.Author).AsMap(),
 	}
 
+	if v, err := lo.FromPtr(t.ConfigItem).TemplateEnv(); err == nil {
```
moshloop, line 89: Shouldn't AsMap just be a call to TemplateEnv() with the removal of fields?

adityathebe: I'm not sure if the fields we were removing should be removed anymore. The `path` could be useful for eg.

`TemplateEnv` could even go away and `AsMap` just returns the map on `.config`

PR #1539, opened 2024-10-22
feat(kratos): use one-time code for account recovery instead of magic links

adityathebe:
one-time codes are the recommended strategy.
https://www.ory.sh/docs/kratos/self-service/flows/account-recovery-password-reset#comparison

"Some email virus scanners open links in emails to scan them. This invalidates the link and may prevent users from completing the flow even if they have the access to the defined address."

### DIFF
--- a/auth/templates.go
+++ b/auth/templates.go
@@ -3,4 +3,6 @@ package auth
 const inviteUserTemplate = `
 <b>Welcome to Mission Control</b>
 <br><br>
-Hello %s, use this link for signing up: %s`
+Hello %s,
+
+please visit <a href="%s">this link</a> to complete registration and use the code: <code>%s</code>`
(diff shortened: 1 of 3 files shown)


auth/templates.go
```diff
@@ -3,4 +3,6 @@ package auth
 const inviteUserTemplate = `
 <b>Welcome to Mission Control</b>
 <br><br>
-Hello %s, use this link for signing up: %s`
+Hello %s,
+
+please visit <a href="%s">this link</a> to complete registration and use the code: <code>%s</code>`
```
moshloop, line 8: Can we switch this to a gotemplate ?

```suggestion
Please visit <a href="%s">%s</a> to complete registration and use the code: <code>%s</code>`
```

PR #1550, opened 2024-10-24
feat: trim down on extraneous calls to ConsumeUntilEmpty in event consumers

adityathebe:
resolves: https://github.com/flanksource/duty/issues/1105

### DIFF
--- a/events/event_queue.go
+++ b/events/event_queue.go
@@ -27,9 +27,7 @@ const (
 	DefaultEventLogSize = 20
 )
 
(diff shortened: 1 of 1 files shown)


events/event_queue.go
```diff
@@ -27,9 +27,7 @@ const (
 	DefaultEventLogSize = 20
 )
 
-type Handler func(ctx context.Context, e models.Event) error
```
adityathebe, line 30: `type Handler` is not required anymore since we merged postq into duty.

PR #1560, opened 2024-10-29
feat: RLS

adityathebe:
merge with: https://github.com/flanksource/duty/pull/1245


auth/tokens.go
```diff
@@ -46,11 +46,22 @@ func GetOrCreateJWTToken(ctx context.Context, user *models.Person, sessionId str
...
 	}
 
-	token, err := jwt.NewWithClaims(jwt.SigningMethodHS256, jwt.MapClaims{
+	claims := jwt.MapClaims{
 		"role": config.Postgrest.DBRole,
 		"id":   user.ID.String(),
-	}).SignedString([]byte(config.Postgrest.JWTSecret))
+	}
+
+	if ctx.Agent() != nil {
```
moshloop, line 54: We need to update the permissions table/model to add agent and tags columns and then call GetImplicitPermissionsForUser,   We could map back to the permission table with v4/5 so that we don't need to parse the expression to get the agent/tag values.

We can't query permissions directly due to roles/groups and deny permissions

auth/tokens.go
```diff
@@ -38,6 +40,60 @@ func InjectToken(ctx context.Context, c echo.Context, user *models.Person, sessI
 	return nil
 }
 
+type RLSPayload struct {
+	Tags   []map[string]string `json:"tags,omitempty"`
+	Agents []string            `json:"agents,omitempty"`
+}
+
+func GetAgentsAndTagPermission(ctx context.Context) (*RLSPayload, error) {
+	permissions, err := rbac.PermsForUser(ctx.User().ID.String())
```
moshloop, line 49: Can we not use GetImplicitPermissions here?

adityathebe: `rbac.PermsForUser` calls `enforcer.GetImplicitPermissionsForUser`

rbac/middleware.go
```diff
@@ -113,6 +113,7 @@ func CheckContext(ctx context.Context, object, action string) bool {
 		return false
 	}
 
+	// TODO: Everyone with an account is not a viewer. i.e. user role.
```
moshloop, line 116: @adityathebe can we implement the guest role + mapping to Clerk Guest role ?

PR #1589, opened 2024-11-06
feat: connection from scraper

adityathebe:


moshloop:
@adityathebe can we add a test for this ?

PR #1593, opened 2024-11-07
feat: filter on notification silence

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/1590

### DIFF
--- a/db/notifications.go
+++ b/db/notifications.go
@@ -125,14 +125,12 @@ func NotificationSendSummary(ctx context.Context, id string, window time.Duratio
 	return earliest.Time, count, err
 }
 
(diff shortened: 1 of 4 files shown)


db/notifications.go
```diff
@@ -125,7 +125,7 @@ func NotificationSendSummary(ctx context.Context, id string, window time.Duratio
 	return earliest.Time, count, err
 }
 
-func GetMatchingNotificationSilencesCount(ctx context.Context, resources models.NotificationSilenceResource) (int64, error) {
+func GetMatchingNotificationSilences(ctx context.Context, resources models.NotificationSilenceResource) ([]models.NotificationSilence, error) {
```
moshloop, line 128: I think `	orClauses := ctx.DB().Where("1 = 0")` needs to change where filter is not null

PR #1600, opened 2024-11-08
chore: soft deleted non-required push topology components

yashmehrotra:


push/topology.go
```diff
@@ -69,20 +68,14 @@ func PushTopology(c echo.Context) error {
 		return dutyAPI.WriteError(c, dutyAPI.Errorf(dutyAPI.EINTERNAL, "error saving components: %v", dutydb.ErrorDetails(err)))
 	}
 
-	var idsToDelete []string
-	if err := ctx.DB().Model(&models.Component{}).Select("id").Where("topology_id = ?", data.TopologyID).Where("deleted_at IS NULL").Where("id NOT IN ?", returnedIDs).Find(&idsToDelete).Error; err != nil {
+	// Soft delete component ids which were not created
+	if err := ctx.DB().Model(&models.Component{}).
```
yashmehrotra, line 72: @moshloop Should we just hard delete these components ? Or let cleanup do its thing ?

moshloop: I think its better to let cleanup handle it

yashmehrotra: 👍🏽 then lets merge it

PR #1607, opened 2024-11-11
test: notification wait for

adityathebe:
merge after: https://github.com/flanksource/duty/pull/1191

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -361,7 +361,7 @@ require (
 	sigs.k8s.io/yaml v1.4.0
 )
 
-// replace github.com/flanksource/duty => ../duty
+replace github.com/flanksource/duty => ../duty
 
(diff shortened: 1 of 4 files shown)


notification/events.go
```diff
@@ -306,6 +306,8 @@ func sendNotifications(ctx context.Context, events models.Events) models.Events
 
 			if !isHealthReportable(notif.Events, previousHealth, currentHealth) {
 				ctx.Logger.V(6).Infof("skipping notification[%s] as health change is not reportable", notif.ID)
+				notificationContext.log.WaitForResolved()
```
adityathebe, line 309: we already save a record on `notification_send_history` table.

This will ensure that we finalize that record when the notification is skipped due to wait for evaluation.

Otherwise, we currenly have records in notification_send_history table with empty status & duration = 0 .

PR #1612, opened 2024-11-11
feat: save pending notifications in send history instead of the event queue

adityathebe:


notification/job.go
```diff
@@ -0,0 +1,43 @@
...
+			tx := ctx.DB().Begin()
+			if tx.Error != nil {
+				return nil
+			}
+			defer tx.Rollback()
+
+			var pending []models.NotificationSendHistory
+			if err := tx.Clauses(clause.Locking{Strength: clause.LockingStrengthUpdate, Options: clause.LockingOptionsSkipLocked}).
+				Where("status = ?", models.NotificationStatusPending).
+				Where("delay IS NULL OR created_at + (delay * INTERVAL '1 second' / 1000000000)  <= NOW()").
```
moshloop, line 31: An expression in the WHERE clause left hand side prevents the use of an index

notification/job.go
```diff
@@ -0,0 +1,43 @@
...
+				Order("created_at + (delay * INTERVAL '1 second' / 1000000000)"). // smallest effective send time first
+				Limit(1).                                                         // one at a time; as one notification failure shouldn't affect a previous successful one
```
moshloop, line 33: This will rate limit notifications to once per 30s, this should probably be a continously running background goroutine that sleeps for 1m on empty

notification/job.go
```diff
@@ -0,0 +1,93 @@
...
+		RunNow:     true,
+		Context:    ctx,
+		Singleton:  false,
+		Schedule:   "@every 30s",
+		Fn: func(ctx job.JobRuntime) error {
+			for {
+				done, err := processPendingNotification(ctx.Context)
+				if err != nil {
+					ctx.History.AddErrorf("failed to send pending notification: %v", err)
+					continue
```
moshloop, line 27:
```suggestion
                                          time.Sleep(2 * time.Second) // prevent spinning on db errors 
					continue
```

Issue #1611, opened 2024-11-11
errors in notification silence filters are not surfaced out


adityathebe:
{"time":"2024-11-11T14:17:58.322019304Z","level":"ERROR","msg":"error consuming: failed to add notification.send event for event=config.unhealthy notification=0191ebb4-2b0d-95c5-ceed-3164c7ed44e7: failed to run filter expression(config.name == \"postgresql\" && config_type ==\"Kubernetes::StatefulSet\"): ERROR: <input>:1:32: undeclared reference to 'config_type' (in container '')\n | config.name == \"postgresql\" && config_type ==\"Kubernetes::StatefulSet\"\n | ...............................^: ERROR: <input>:1:32: undeclared reference to 'config_type' (in container '')\n | config.name == \"postgresql\" && config_type ==\"Kubernetes::StatefulSet\"\n | ...............................^"}
{"time":"2024-11-11T14:18:03.346806324Z","level":"ERROR","msg":"error consuming: failed to add notification.send event for event=config.unhealthy notification=0191ebb4-2b0d-95c5-ceed-3164c7ed44e7: failed to run filter expression(config.name == \"postgresql\" && config_type ==\"Kubernetes::StatefulSet\"): ERROR: <input>:1:32: undeclared reference to 'config_type' (in container '')\n | config.name == \"postgresql\" && config_type ==\"Kubernetes::StatefulSet\"\n | ...............................^: ERROR: <input>:1:32: undeclared reference to 'config_type' (in container '')\n | config.name == \"postgresql\" && config_type ==\"Kubernetes::StatefulSet\"\n | ...............................^"}

moshloop:
@adityathebe  - Can we also update the docs of the fields available ?

adityathebe:
> @adityathebe - Can we also update the docs of the fields available ?

https://github.com/flanksource/docs/issues/309
https://github.com/flanksource/flanksource-ui/issues/2403

PR #1615, opened 2024-11-13
feat: errors in notification silences

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/1611

### DIFF
--- a/db/notifications.go
+++ b/db/notifications.go
@@ -100,6 +100,10 @@ func UpdateNotificationError(ctx context.Context, id string, err string) error {
 	return ctx.DB().Model(&models.Notification{}).Where("id = ?", id).Update("error", err).Error
 }
 
+func UpdateNotificationSilenceError(ctx context.Context, id string, err string) error {
+	return ctx.DB().Model(&models.NotificationSilence{}).Where("id = ?", id).Update("error", err).Error
+}
+
 func DeleteNotificationSendHistory(ctx context.Context, days int) (int64, error) {
 	tx := ctx.DB().
 		Model(&models.NotificationSendHistory{}).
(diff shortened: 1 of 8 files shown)


notification/events.go
```diff
@@ -547,18 +545,20 @@ func GetEnvForEvent(ctx context.Context, event models.Event) (*celVariables, err
...
 }
 
-func shouldSilence(celEnv map[string]any, matchingSilences []models.NotificationSilence) (bool, error) {
+func shouldSilence(ctx context.Context, celEnv map[string]any, matchingSilences []models.NotificationSilence) bool {
 	for _, silence := range matchingSilences {
 		if silence.Filter != "" {
 			if ok, err := gomplate.RunTemplateBool(celEnv, gomplate.Template{Expression: string(silence.Filter)}); err != nil {
-				return false, fmt.Errorf("failed to run filter expression(%s): %w", silence.Filter, err)
+				ctx.Errorf("failed to run silence filter expression(%s): %v", silence.Filter, err)
+				logs.IfError(db.UpdateNotificationSilenceError(ctx, silence.ID.String(), err.Error()), "failed to update notification silence")
```
moshloop, line 553: this is a type safe update, so that the draft mode and module replacement are not necessary - on open the build will fail, and then the PR can just be rebased to make it pass

```suggestion
				logs.IfError(ctx.DB().Model(&filter).Select("error").Update(models.NotificationSilence{Error: lo.ToPtr(err.Error())}), "failed to update notification silence")
```

db/notifications.go
```diff
@@ -167,7 +171,7 @@ func GetMatchingNotificationSilences(ctx context.Context, resources models.Notif
 	query = query.Where(orClauses)
 
 	var silences []models.NotificationSilence
-	err := query.Where(`"from" <= NOW()`).Where("until >= NOW()").Where("deleted_at IS NULL").Find(&silences).Error
+	err := query.Where(`"from" <= NOW()`).Where("error IS NULL").Where("until >= NOW()").Where("deleted_at IS NULL").Find(&silences).Error
```
moshloop, line 174: We might need to null out error, when the filter is updated as otherwise the silence is ignored forever

Issue #1339, opened 2024-09-03
Notification Silences 2.0


moshloop:
- [x] 1. CRD e.g. like https://github.com/giantswarm/silence-operator

PR #1688, opened 2024-12-20
feat: notification silence operator

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/1339


fixtures/notifications/silence-test-env.yaml
```diff
@@ -0,0 +1,11 @@
+---
+apiVersion: mission-control.flanksource.com/v1
+kind: NotificationSilence
+metadata:
+  name: test-env-silence
+spec:
+  from: 2024-01-01
+  until: 2025-01-01
```
adityathebe, line 8: @moshloop we probably should make these fields optional.

Issue #1538, opened 2024-10-22
Playbooks - Secret parameters


moshloop:
- Save encrypted values to the DB
- Don't return the values on any API/DB call

moshloop:
See https://pkg.go.dev/gocloud.dev@v0.40.0/secrets

PR #1775, opened 2025-01-20
feat: playbook secret parameters

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/1538


playbook/runner/template.go
```diff
@@ -64,6 +72,29 @@ func CreateTemplateEnv(ctx context.Context, playbook *models.Playbook, run *mode
 				templateEnv.Params[p.Name] = component.AsMap()
 			}
 
+		case v1.PlaybookParameterTypeSecret:
```
moshloop, line 75: We need to ensure that errors / logs / oops do not include secret values - Maybe we create a SanitizingLogger and pass it a map of decrypted values and set it on the context, GetLoggingContext could also use a sanitizer so that errors don't include it  

This should probably be in duty as `duty.EncryptValue` `duty.DecryptValue` which returns a `SecureString` - which has a custom JSONMarshall/String methods that return "*redacted*" instead of the secret value which is accessible by `SecureString.GetValue()`.

adityathebe: https://github.com/flanksource/duty/pull/1284

Issue #1354, opened 2024-09-06
Notifications - Custom Grouping 


flankbot:

moshloop:
i.e When tenant upgrades fail we should receive a single notification and a list of resources affected, Mark the duplicates as "grouped"

PR #1781, opened 2025-01-23
feat: grouped notifications

yashmehrotra:
Fixes: #1354


notification/job.go
```diff
@@ -259,23 +266,65 @@ func processPendingNotification(ctx context.Context, currentHistory models.Notif
...
+	if len(historiesToUpdate) > 1 {
+		payload.GroupedResources = historiesToUpdate[1:]
+	}
+	historyIDs := lo.Map(historiesToUpdate, func(h models.NotificationSendHistory, _ int) uuid.UUID { return h.ID })
+
+	if err := sendPendingNotification(ctx, historyToUpdate, payload); err != nil {
 		return fmt.Errorf("failed to send notification: %w", err)
-	} else if dberr := ctx.DB().Model(&models.NotificationSendHistory{}).Where("id = ?", currentHistory.ID).UpdateColumns(map[string]any{
+	} else if dberr := ctx.DB().Model(&models.NotificationSendHistory{}).Where("id IN ?", historyIDs).UpdateColumns(map[string]any{
 		"status": models.NotificationStatusSent,
```
moshloop, line 328: Shouldn't this be grouped ?

notification/job.go
```diff
@@ -189,7 +189,21 @@ func ProcessPendingNotifications(parentCtx context.Context) (bool, error) {
 			currentHistory.ResourceID,
 		)
 
-		if err := processPendingNotification(ctx, currentHistory); err != nil {
+		var groupedHistory []models.NotificationSendHistory
```
moshloop, line 192: I think this is prone to race conditions - We should probably have a different job to group the notifications by groupByPeriod i.e. for each unique hash, `UPDATE send_history set status = Grouped where id in (select id from history where hash=1 and created_at < now-group by period, order by created_at offset 1)`

and then waitFor = max(waitFor, groupByPeriod + 60s)

We also need to update the indexes to match these queruies

PR #1787, opened 2025-01-23
feat: template env improvements

adityathebe:
resolves: https://github.com/flanksource/duty/issues/1286


db/common.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"time"
+
+	"github.com/flanksource/duty/context"
+	"github.com/patrickmn/go-cache"
+)
+
+var distinctTagsCache = cache.New(time.Minute*10, time.Hour)
+
+func GetDistinctTags(ctx context.Context) ([]string, error) {
```
moshloop, line 12: shouldn't this be part of duty ?

notification/cel.go
```diff
@@ -98,6 +102,60 @@ func (t *celVariables) AsMap() map[string]any {
 		output["new_state"] = t.NewState
 	}
 
+	// Placeholders for commonly used fields of the resource
```
moshloop, line 105: this seems to be duplicated, can we just have a duty.GetResourceContext(selectable) ?

Issue #1789, opened 2025-01-24
Unable to delete notifications


adityathebe:
{level: ERROR, logger: db, msg: ERROR >=[33m[3ms] [34;1m[rows:0][0m ERROR: update or delete on table "notification_send_history" violates foreign key constraint "playbook_run_notification_send_id_fkey" on table "playbook_runs" (SQLSTATE 23503) DELETE FROM "notifications" WHERE id = $1$}

PR #1792, opened 2025-01-24
fix: soft delete on CRD delete

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/1789

ran into this when deleting a notification crd got stuck.
notification_send_history had foreign key to playbook_runs.


moshloop:
How does this handle delete and recreate ? Are these deleted items ever cleaned up?

PR #1821, opened 2025-02-04
feat: allow wildcard on subject.people

adityathebe:
* allow specifying the subject group directly


api/v1/permission_types.go
```diff
@@ -50,36 +49,39 @@ func (t PermissionSubjectSelector) Find(ctx context.Context, table string) (stri
...
-
-		case 2:
-			namespace, name := splits[0], splits[1]
-			var id string
-			err := ctx.DB().Select("id").Table(table).
-				Where("namespace = ?", namespace).
-				Where("name = ?", name).
-				Find(&id).Error
-			return id, models.PermissionSubjectTypeNotification, err
+		if len(splits) != 2 {
```
moshloop, line 52: I think we should check if t is a valid uuid, and if so use that across all table types.

PR #1822, opened 2025-02-04
feat: matchQuery cel function

adityathebe:


playbook/events.go
```diff
@@ -239,7 +239,7 @@ func matchResource(ctx context.Context, labels map[string]string, celEnv map[str
 outer:
 	for _, mf := range matchFilters {
 		if mf.Filter != "" {
-			res, err := ctx.RunTemplate(gomplate.Template{Expression: mf.Filter}, celEnv)
+			res, err := ctx.RunTemplate(gomplate.Template{CelEnvs: utils.CelFunctions, Expression: mf.Filter}, celEnv)
```
moshloop, line 242: shouldn't this be included in all celenvs by default ?

moshloop: Or atleast in duty run.template

adityathebe: ya. for some reason I thought `grammar` was in `flanksource/mission-control`

PR #1831, opened 2025-02-10
feat: permission for connection

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/638

### DIFF
--- a/api/notifications.go
+++ b/api/notifications.go
@@ -4,7 +4,6 @@ import (
 	gocontext "context"
 	"database/sql/driver"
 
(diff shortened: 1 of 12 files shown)


moshloop (changes requested):
I think we need to add permissions.connections to the chart to make connection permissions opt-in + update the values schema

adityathebe:
https://github.com/flanksource/mission-control-chart/pull/235

@moshloop

moshloop (approved):

PR #1842, opened 2025-02-13
chore: add updated_at in notification trace log

yashmehrotra:


yashmehrotra:
@moshloop Self merging FYI

PR #1847, opened 2025-02-14
chore: shutdown and exit in fatal scenarios

yashmehrotra:
@moshloop @adityathebe Should be put the shutdown and exit in logger.Fatalf ? or keep it the way it is right now ?


adityathebe:
I usually directly use `shutdown.ShutdownAndExit` but either is fine with me 👍

Issue #1846, opened 2025-02-13
Playbook actions retry options


moshloop:
To handle rate limiting 

```
retry:
  limit: 5
  duration: 30s # period between retries
  jitter: 10% (duration +-
  exponential:
    multiplier: 1.5
```

Each action should be recorded individually

PR #1851, opened 2025-02-17
feat: Playbook action retries

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/1846


playbook/runner/functions.go
```diff
@@ -49,6 +49,11 @@ func GetActionByName(ctx context.Context, runID, actionName string) (map[string]
 
 	var action models.PlaybookRunAction
 	query := ctx.DB().Where("name = ?", actionName).Where("playbook_run_id = ?", runID)
+
+	// there could be multiple actions with the same name due to retries.
+	// we fetch the latest one
+	query.Order("start_time ASC").Limit(1)
```
moshloop, line 55: Would this be the first one ?

api/v1/playbook_actions.go
```diff
@@ -509,6 +511,42 @@ func (t *AWSConnection) Populate(ctx connectionContext, k8s kubernetes.Interface
...
+	Limit int `json:"limit"`
+
+	// Duration is the duration to wait before retrying the action.
+	Duration string `json:"duration"`
+
+	// Jitter is the random factor to apply to the duration.
+	// Ranges from 0 to 100.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=100
+	Jitter float64 `json:"jitter,omitempty"`
```
moshloop, line 530: Shouldn't this be `int` then ?

moshloop (changes requested):
Can we remove the need for allowDangerousTypes

moshloop (approved):

PR #1853, opened 2025-02-18
fix: use waitForGroup in CRD

adityathebe:
repeatGroup (crd) was mapped to GroupByHash.

RepeatGroup was originally meant for different purpose.
Removing it for now as we aren't using it anywhere.

### DIFF
--- a/api/v1/notification_types.go
+++ b/api/v1/notification_types.go
@@ -77,9 +77,9 @@ type NotificationSpec struct {
 	// Kubernetes config health. Format: "5m", "1h"
 	WaitForEvalPeriod *string `json:"waitForEvalPeriod,omitempty" yaml:"waitForEvalPeriod,omitempty"`
 
-	// WaitForGroup allows notifications in waiting status to be grouped together
+	// GroupBy allows notifications in waiting status to be grouped together
 	// based on certain set of keys.
-	WaitForGroup []string `json:"waitForGroup,omitempty"`
+	GroupBy []string `json:"groupBy,omitempty"`
 }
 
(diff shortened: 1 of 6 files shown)


api/v1/notification_types.go
```diff
@@ -76,6 +76,10 @@ type NotificationSpec struct {
 	// WaitForEvalPeriod is an additional delay after WaitFor before evaluating
 	// Kubernetes config health. Format: "5m", "1h"
 	WaitForEvalPeriod *string `json:"waitForEvalPeriod,omitempty" yaml:"waitForEvalPeriod,omitempty"`
+
+	// WaitForGroup allows notifications in waiting status to be grouped together
+	// based on certain set of keys.
+	WaitForGroup []string `json:"waitForGroup,omitempty"`
```
moshloop, line 82: This should probably be groupBy with an error thrown if groupBy is specified without a waitFor

api/v1/notification_types.go
```diff
@@ -76,6 +76,10 @@ type NotificationSpec struct {
 	// WaitForEvalPeriod is an additional delay after WaitFor before evaluating
 	// Kubernetes config health. Format: "5m", "1h"
 	WaitForEvalPeriod *string `json:"waitForEvalPeriod,omitempty" yaml:"waitForEvalPeriod,omitempty"`
+
+	// GroupBy allows notifications in waiting status to be grouped together
+	// based on certain set of keys.
```
moshloop, line 81: Should provide examples of these keys, as well as the diff/hash key

Issue #1858, opened 2025-02-20
Notification Fallback Connection


moshloop:
fallback:
   connection|  etc..
   timeout: # wait this long before considering a send a failure

PR #1867, opened 2025-02-24
feat: notification fallback recipients

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/1858


notification/job.go
```diff
@@ -303,6 +303,187 @@ func ProcessPendingGroupedNotifications(parentCtx context.Context) error {
...
+					break
+				}
+			}
+
+			return nil
+		},
+	}
+}
+
+func ProcessFallbackCheckNotifications(parentCtx context.Context) (bool, error) {
```
moshloop, line 344: Why do we need jobs for this? Can't we do it inline

adityathebe: https://github.com/flanksource/duty/pull/1347

PR #1877, opened 2025-02-27
chore: set system user id

yashmehrotra:


db/people.go
```diff
@@ -121,3 +121,16 @@ func AddPersonToTeam(ctx context.Context, personID uuid.UUID, teamID uuid.UUID)
...
+var SystemUser *models.Person
+
+func GetSystemUser(ctx context.Context) (*models.Person, error) {
+	if SystemUser == nil {
+		if err := ctx.DB().Model(&models.Person{}).Where("name = 'System'").First(&SystemUser).Error; err != nil {
+			return nil, fmt.Errorf("error fetching system user from database: %w", err)
+		}
+	}
+
+	api.SystemUserID = lo.ToPtr(SystemUser.ID)
```
yashmehrotra, line 134: @moshloop Keeping api.SystemUserID to avoid import cycles,  its used in a lot of structs defined in api package

PR #1922, opened 2025-03-10
fix: handle user induced SQL errors when running playbooks

adityathebe:


adityathebe:
Not sure if we should even be running the action with an open transaction.

playbook/run_consumer.go
```diff
@@ -271,7 +272,25 @@ func ActionConsumer(ctx context.Context) (int, error) {
 			return err
 		}
 
-		if err := runner.RunAction(ctx, run, action); err != nil {
+		// Create a SAVEPOINT to prevent full transaction rollback
+		if err := ctx.DB().Exec("SAVEPOINT action_execution").Error; err != nil {
```
moshloop, line 276: Can we create a test for this ?

Issue #1871, opened 2025-02-25
AI - Use structured JSON output


moshloop:
Instead of asking the AI to build the slack message, we ask it to produce structured json matching a schema:

`
{
title: 
summary: 
severity: 
events: [
],
playbooks: [
  id: abc
  params: {
 "a": b,
 }
],
links: [
{ name: "Grafana",
  url: ..  
}
}
`

We can then have a function to turn the AI response into a slack block, and another function to turn a config item/comonent/canary into the same object structure and then have a merge eg.

```golang
slackMerge(configItemToSlack(.config), aiToSlack(.actions.ai.output))
```

wdyt @adityathebe @yashmehrotra

PR #1938, opened 2025-03-13
feat(llm): structure output for diagnosis & recommendations

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/1871

- Use response_format for OpenAI
- For all other LLMS, use tools to get structured output
  - Use official gemini sdk for structured output because langchaingo didn't offer a way to force the tool use
  - Langchaingo also doesn't allow forcing tool use for anthropic but we force that with a prompt which seems to work fine. Waiting for langchaingo to allow that.


llm/llm.go
```diff
@@ -48,28 +39,56 @@ func Prompt(ctx context.Context, config Config, systemPrompt string, promptParts
...
+	// Add backend-specific options for tool choice
+	switch config.Backend {
+	case api.LLMBackendOpenAI:
+		// Do nothing
+		// NOTE: we use `response_format` instead of function calling & that's configured during model creation not when prompting.
+		// OpenAI does support function calling, but I don't think we can force the model to use that tool
+		// like we can with Anthropic.
+
+	case api.LLMBackendGemini:
+		// Do nothing
```
moshloop, line 64: According to https://pkg.go.dev/google.golang.org/genai#GenerateContentConfig there is 
```
// Schema that the generated candidate text must adhere to.
	ResponseSchema *[Schema](https://pkg.go.dev/google.golang.org/genai#Schema) `json:"responseSchema,omitempty"`
```

adityathebe: hmm, we are using https://github.com/google/generative-ai-go/genai
which doesn't seem to have that config. We do it via tools.

But googleapis/go-genai seems to be the new repo for it (based on the commits)
Will migrate to that.

adityathebe: https://github.com/flanksource/mission-control/pull/1938/files#diff-cfdf50fee4d1c92bc2e7bfb32ce2c3fa7b2f53f4a8d0c95b9d6883e4b8c8c535R140-R148

PR #1963, opened 2025-03-21
feat: save AI prompts to artifacts

adityathebe:
related: https://github.com/flanksource/mission-control/issues/1961

### DIFF
--- a/playbook/actions/ai.go
+++ b/playbook/actions/ai.go
@@ -5,10 +5,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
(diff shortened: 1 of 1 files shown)


moshloop:
@adityathebe can we check what happens when no artificat store is configured ? - It should just be a warning in the logs

adityathebe:
@moshloop that's how it is right now - https://github.com/flanksource/mission-control/blob/8c70e3d7cfd2b68a304e0f6fbf7cfe9e1c40dd87/playbook/runner/artifacts.go#L22-L25

PR #1996, opened 2025-04-01
chore: add header support for smtp

yashmehrotra:


utils/map.go
```diff
@@ -42,3 +44,28 @@ func (m *SyncedMap[K, V]) Append(key K, value V) {
...
+		return m, nil
+	}
+
+	err := json.Unmarshal([]byte(s), &m)
+	if err != nil {
+		return nil, fmt.Errorf("error unmarshaling string: %w", err)
+	}
+
+	return m, nil
+}
```
adityathebe, line 71: You could use `collections.SelectorToMap` and `collections.SortedMap`

yashmehrotra: no func to convert selector string to map though

yashmehrotra: `,` is allowed in headers and that would mess with parsing

Issue #1987, opened 2025-03-31
Child playbook runs are not created for *some* runs


adityathebe:
This is causing the runs to time out

PR #1997, opened 2025-04-01
fix: improve playbook permissions based on initiation source

adityathebe:
- Use initiating user as permission subject for user-triggered playbooks
- Use source notification as permission subject for notification-triggered playbooks
- Use source playbook as permission subject for playbook-triggered playbooks
- 
resolves: https://github.com/flanksource/mission-control/issues/1987


moshloop:
One of the use-cases for playbooks is to provide guardrails, if the playbook needs the same permissions as the user then it wouldn't be possible for example to grant a playbook access to a connection so a user can perform a limited action

PR #1998, opened 2025-04-01
chore: simplify http playbook output result

yashmehrotra:


playbook/actions/http.go
```diff
@@ -11,16 +11,12 @@ import (
 	v1 "github.com/flanksource/incident-commander/api/v1"
 )
 
-type HTTPResultResponse struct {
+type HTTPResult struct {
+	Content    string            `json:"content"`
 	Headers    map[string]string `json:"headers"`
 	StatusCode int               `json:"statusCode"`
```
moshloop, line 17: to make it consistent with http check
```suggestion
	StatusCode int               `json:"code"`
```

PR #2011, opened 2025-04-04
feat: inhibition by grouping parameters

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/2007


notification/events.go
```diff
@@ -324,6 +321,67 @@ func checkInhibition(ctx context.Context, notif NotificationWithSpec, resource t
...
+					hash, err := calculateGroupByHash(ctx, inhibition.GroupBy, sendHistory.ResourceID.String(), payload.EventName)
+					if err != nil {
+						return nil, fmt.Errorf("failed to calculate group by hash: %w", err)
+					}
+
+					if hash == requiredHash {
+						return lo.ToPtr(sendHistory.ID), nil
+					}
+				}
+			}
```
adityathebe, line 379: We either

1. precalculate the hash for all inhibition rules and save it in the send history (like we do with group_by_hash). The problem with that is
- Doesn't work if inhibition rule changes
- we'd have to store 1 groupByHash per inhibition rule (maybe in a array column). But, w'd need to id which inhibition rule a groupByHash in the list is for. So the inhibition rule would need a name also?

2. Calculate the hash on-demand. Flexible with changing inhibition rules but inefficient as the number of inhibition rules/window grows

PR #2027, opened 2025-04-09
feat: remove resources from group once resolved

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/2021


notification/events.go
```diff
@@ -158,6 +164,54 @@ func (t *notificationHandler) addNotificationEvent(ctx context.Context, event mo
 	return nil
 }
 
+// resolveGroupMembership removes any resources from notification group that
```
moshloop, line 167: I think it better to track lastResolvedAt and status = Resolved

Issue #2044, opened 2025-04-17
AI action - Child Runs: We only feed action.result to the model. Artifacts from those actions also need to be supplied


adityathebe:
Logs playbook action will store logs as artifacts

PR #2055, opened 2025-04-22
feat: playbook artifacts as context

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/2044


artifacts/artifacts.go
```diff
@@ -16,6 +20,114 @@ import (
...
+	var contents []ArtifactContent
+	for _, a := range actionArtifacts {
+		reader, err := fs.Read(ctx, a.Path)
+		if err != nil {
+			return nil, fmt.Errorf("error loading artifact: %w", err)
+		}
+		defer reader.Close()
+
+		if maxSize > 0 {
+			reader = io.NopCloser(io.LimitReader(reader, maxSize))
```
moshloop, line 114: I think this should be a tail and not a head (and rename function to TailArtificats)

adityathebe: Good point. 

But since artifactFS doesn't natively support this, does it make sense to tail since we'd need to read the entire body and then truncate?

The `LimitReader` here can act as a safeguard that ideally we will never need. The caller can then tail or read all.

Issue #2065, opened 2025-04-24
Notifications - Add AI playbook summary


moshloop:
once the playbook completes look for an action that includes a summary OR add a summary/title the playbook_run

PR #2081, opened 2025-05-01
feat: save AI action results to notification body & title

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/2065

### DIFF
--- a/playbook/runner/runner.go
+++ b/playbook/runner/runner.go
@@ -178,6 +178,70 @@ func getEligibleAgents(spec v1.PlaybookSpec, action *v1.PlaybookAction, run mode
 	return []string{Main}
 }
 
+func saveAIResultToSendHistory(ctx context.Context, run models.PlaybookRun) error {
+	if run.NotificationSendID == nil {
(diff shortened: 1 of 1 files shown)


playbook/runner/runner.go
```diff
@@ -178,6 +178,70 @@ func getEligibleAgents(spec v1.PlaybookSpec, action *v1.PlaybookAction, run mode
...
+		if lo.Contains(notificationActionNames, action.Name) {
+			if slackMsg, ok := action.Result["slack"].(string); ok {
+				sendHistoryUpdate.Body = &slackMsg
+			} else if body, ok := action.Result["body"].(string); ok {
+				sendHistoryUpdate.Body = &body
+			}
+		}
+	}
+
+	if len(sendHistoryUpdate.ResourceHealthDescription) > 0 {
```
moshloop, line 236: What happens if there is just a notification in the playbook?

adityathebe: We save the body sent by the notification action to the send history.

PR #2139, opened 2025-06-11
fix: GCP application backups & regions

adityathebe:
pull in locations and database backups from GCP

image

resolves: https://github.com/flanksource/mission-control/issues/2138


moshloop:
I think the account should be the name and "prod" the tag

PR #2141, opened 2025-06-12
feat: application changes

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/2140

### DIFF
--- a/api/application.go
+++ b/api/application.go
@@ -77,8 +77,6 @@ type ApplicationChange struct {
 	User        string    `json:"user"`
 	Description string    `json:"description"`
 	Status      string    `json:"status"`
-	CreatedAt   time.Time `json:"createdAt"`
-	UpdatedAt   time.Time `json:"updatedAt"`
 }
 
 type ApplicationAccessControl struct {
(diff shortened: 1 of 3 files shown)


moshloop:
I think there are 2 types of changes - config changes on existing items, and also config items themeselves which could have status / approvers / etc...

1. Audit logs (i.e. User was given access, user access was removed) which would be group by config item and queries based on change type.
```yaml
spec:
   environments:
      - name: Production
         tags: env=prod,account=Production # list of tag filters to identify environment for incidents/etc.
    changes:
       - query: type=MissionControl::Application 
          # where in the report to list the changes
          section: Users # Databases, Releases, Deployments, Incidents, 
          mapping:
               - query: changeType=AccessGranted,AccessRevoked
                  status: Success
                - query: changeType=AccessDenied
                   status: Denied
           
       -  query: type=Jira::Issue env=prod # we only care about prod incidents
           section: Incidents
           # no status mapping, so use is-healthy for status

       -  query: type= Github::ActionRun
           section: Deployments
           mapping:
               - approver:  .config.jobs.steps.approve.approver_id
           # no status mapping, so use is-healthy for status

```

Issue #2145, opened 2025-06-13
Add functionality to generate a new token for an existing agent


yashmehrotra:

adityathebe:
related: https://github.com/flanksource/mission-control/issues/1243

Issue #1243, opened 2024-08-08
Auto renew tokens 


moshloop:
Add a autoRenew duration - default to 7 days

PR #2146, opened 2025-06-13
feat: add option to generate new token for existing agent

yashmehrotra:
Fixes: #2145 
Fixes: https://github.com/flanksource/mission-control/issues/1243

### DIFF
--- a/auth/tokens.go
+++ b/auth/tokens.go
@@ -233,14 +233,14 @@ func getAccessToken(ctx context.Context, token string) (*models.AccessToken, err
 	if expiry == nil {
 		tokenCache.Set(token, &accessToken, -1)
 	} else {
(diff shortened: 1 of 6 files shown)


auth/tokens.go
```diff
@@ -222,14 +229,21 @@ func getAccessToken(ctx context.Context, token string) (*models.AccessToken, err
 		return nil, err
 	}
 
-	if accessToken.ExpiresAt == nil {
+	expiry := accessToken.ExpiresAt
+	if expiry == nil {
 		tokenCache.Set(token, &accessToken, -1)
 	} else {
-		if accessToken.ExpiresAt.Before(time.Now()) {
+		if time.Until(*expiry) < preExpiryWindow {
```
moshloop, line 236: This should probably happen after the expiry check

PR #2173, opened 2025-06-27
feat: create tables from views

adityathebe:


jobs/views.go
```diff
@@ -0,0 +1,49 @@
...
+				viewResource, err := v1.ViewFromModel(&view)
+				if err != nil {
+					return fmt.Errorf("failed to convert view model to resource: %v", err)
+				}
+
+				if _, err := views.PopulateView(ctx.Context, viewResource); err != nil {
+					return fmt.Errorf("failed to populate view %s/%s: %v", view.Namespace, view.Name, err)
+				}
+
+				ctx.History.IncrSuccess()
```
moshloop, line 42: this should be the number of rows in the view

api/v1/view_types.go
```diff
@@ -38,25 +47,32 @@ type ViewSpec struct {
 
 	// Queries define the queries and mappings to populate the view
 	Queries ViewQueriesSpec `json:"queries" yaml:"queries"`
+
+	// CacheTTL defines how long to cache the view data (e.g., "1h", "30m")
+	CacheTTL string `json:"cacheTTL,omitempty" yaml:"cacheTTL,omitempty"`
```
moshloop, line 52:
```suggestion
type Cache struct {
    // The maximum age to cache views for, defaults to "15m"
	MaxAge string 
	//  The minumum time between changes, in the underlying view data, the query will be run at most every MinAge, defaults to "10s"
	MinAge string
	// The timeout in milliseconds for a refresh query to execute, before returning stale data, defaults to 5000
	RefreshTimeout int  
	`json:"cacheTTL,omitempty" yaml:"cacheTTL,omitempty"`
```

views/table.go
```diff
@@ -0,0 +1,108 @@
...
+
+	var finalPanelResults []api.PanelResult
+	if len(panelResult.Results) > 0 {
+		if err := json.Unmarshal(panelResult.Results, &finalPanelResults); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal panel results: %w", err)
+		}
+	}
+
+	return &api.ViewResult{
+		Columns: view.Spec.Columns,
```
moshloop, line 54: Maybe add last refreshed at

PR #2189, opened 2025-07-02
fix(view): gauge column type

adityathebe:
related: https://github.com/flanksource/mission-control/issues/2162


db/views_table.go
```diff
@@ -127,5 +130,22 @@ func ReadViewTable(ctx context.Context, tableName string) ([]api.ViewRow, error)
...
 
-	return viewRows, nil
+	return postProcessViewRows(viewRows, columnDef), nil
+}
+
+func postProcessViewRows(viewRows []api.ViewRow, columnDef []api.ViewColumnDef) []api.ViewRow {
+	for _, viewRow := range viewRows {
+		for i, colDef := range columnDef {
+			switch colDef.Type {
+			case api.ViewColumnTypeGauge:
```
moshloop, line 140: I think we might need more safety in type conversions here

db/views_table.go
```diff
@@ -127,5 +131,57 @@ func ReadViewTable(ctx context.Context, tableName string) ([]api.ViewRow, error)
...
+				case int:
+					viewRow[i] = time.Duration(v)
+				case int32:
+					viewRow[i] = time.Duration(v)
+				case int64:
+					viewRow[i] = time.Duration(v)
+				case float64:
+					viewRow[i] = time.Duration(int64(v))
+				default:
+					logger.Warnf("postProcessViewRows: unknown duration type: %T", v)
```
moshloop, line 166: I think we should return these as errors - they will pollute the logs and hide it from the user

adityathebe:
Already added to duty and missioncontrol in different PR

Issue #2201, opened 2025-07-09
Notifications - Improve messaging


moshloop:
1. Only show whitelist labels in notifications: e.g.

notifications.whitelist.labels=app|batch.kubernetes.io/jobname|app.kubernetes.io/name;app.kubernetes.io/version|

Where groups are split by `;` and only the first label with a value is shown

2. Remove labels header

PR #2212, opened 2025-07-11
fix: notification msg improvements

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/2201


playbook/actions/ai_slack.go
```diff
@@ -205,9 +206,12 @@ func slackBlocks(ctx context.Context, knowledge *KnowledgeGraph, diagnosisReport
 
 	blocks = append(blocks, divider)
 
-	if labelsSection := createSlackFieldsSection("Labels", *affectedResource.Labels); labelsSection != nil {
-		blocks = append(blocks, labelsSection)
-		blocks = append(blocks, divider)
+	if affectedResource.Labels != nil {
```
moshloop, line 209: I don't think this should be specific to AI notifications, rather their should be a generic configItem.getLabels() []Label that returns a sorted list of labels, starting with tags (filtering out common values e.g. if both `eu-west-1` and `eu-west-1a` tag values are found only return the more specific `eu-west-1a` zone, and then whitelisted labels.

getLabels should then be called by normal notifications / ai / etc.. 

For both tags and labels use custom weights:

* account
* project
* cluster
* namespace
* name
* service
* deployment
* statefulset
* daemonset
* cronjob
* pod
* job
* container
* release
* chart

then A-Z

then negative weighting (if whitelisted)
* version
* *-hash
* *generation

PR #2246, opened 2025-07-29
chore: push agent as config item to upstream

yashmehrotra:


jobs/upstream.go
```diff
@@ -32,6 +38,35 @@ func ReconcileAllJob(config upstream.UpstreamConfig) *job.Job {
...
+				if ctx.History.SuccessCount > 0 {
+					health = dutymodels.HealthWarning
+				} else if ctx.History.SuccessCount == 0 {
+					health = dutymodels.HealthUnhealthy
+				}
+			}
+
+			err := client.Push(ctx.Context, &upstream.PushData{
+				ConfigItems: []dutymodels.ConfigItem{{
+					ID:          uuid.New(),
```
yashmehrotra, line 55: Add comment that it gets overridden by acutal agen id in push handler

PR #2305, opened 2025-08-28
feat: add token permissions

yashmehrotra:


moshloop:
@yashmehrotra can we add some tests ?

PR #2312, opened 2025-09-03
fix: update description to fix helm install error

yashmehrotra:


yashmehrotra:
@moshloop Self merging FYI

PR #2337, opened 2025-09-19
fix(notification): notification properties should overwrite connection properties

yashmehrotra:


notification/send.go
```diff
@@ -221,7 +221,7 @@ func SendNotification(ctx *Context, connectionName, shoutrrrURL string, celEnv m
 		ctx.WithRecipient(RecipientTypeConnection, &connection.ID)
 
 		shoutrrrURL = connection.URL
-		data.Properties = collections.MergeMap(connection.Properties, data.Properties)
```
adityathebe, line 224: Isn't it already like that?

yashmehrotra: My bad, problem was with using multiple options with to, conn props had `to` and playbook had `ToAddresses` and we give precendence to `to`, updating playbook

Also, I think we should only use 1 keyword instead of multiple with fallback

PR #2360, opened 2025-09-29
feat(RLS): use middleware from duty

adityathebe:
dependsOn: https://github.com/flanksource/duty/pull/1563


adityathebe:
No need to migrate to duty.

PR #2388, opened 2025-10-06
feat: RLS with name support

adityathebe:
[skip ci]


adityathebe:
we have access scope now

Issue #2407, opened 2025-10-14
Playbooks | Show forbidden error instead of displaying no run parameter


adityathebe:
when a guest user that has `read` permission to a playbook but not a `playbook:run` permision opens the run modal, they'll not see the parameters. 

This is because they don't have the permission to run. But this is not a good UX.

PR #2414, opened 2025-10-15
feat: /can-i endpoint

adityathebe:
related: https://github.com/flanksource/mission-control/issues/2407


adityathebe:
Don't need it jsut yet

PR #2482, opened 2025-11-03
Fix access token deletion cleanup

adityathebe:
## Summary
Ensures that when an access token is deleted, all associated RBAC roles and permissions are also removed from the system, and the token's person record is soft deleted.

resolves: https://github.com/flanksource/mission-control/issues/2466


moshloop:
@adityathebe  Why should deleting an access token delete the user/permissions?

PR #2493, opened 2025-11-06
MCP auth middleware

adityathebe:
Remove mcp check action=* on mcp object.

we allow mcp access, if the user has mcp:run access on at least one object.


moshloop:
@adityathebe can you resolve conflicts ?

PR #2526, opened 2025-11-14
feat(MCP): Session specific tools

adityathebe:
- [x] Only create playbook -> tools if the user token can read the playbook.
- [x] remove the playbook registry and the syncPlaybook job
- [x] fix playbook run DB transaction issue

### DIFF
--- a/mcp/suite_test.go
+++ b/mcp/suite_test.go
@@ -4,15 +4,14 @@ import (
 	"net/http/httptest"
 	"testing"
 
+	"github.com/flanksource/duty/context"
+	"github.com/flanksource/duty/tests/setup"
 	echov4 "github.com/labstack/echo/v4"
 	"github.com/mark3labs/mcp-go/client"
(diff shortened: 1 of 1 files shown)


mcp/playbooks.go
```diff
@@ -27,6 +26,53 @@ import (
...
+
+const (
+	toolGetPlaybookRunSteps   = "get_playbook_run_steps"
+	toolGetFailedPlaybookRuns = "playbook_failed_runs"
+	toolGetRecentPlaybookRuns = "playbook_recent_runs"
+	toolPlaybooksListAll      = "playbooks_list_all"
+)
+
+func addPlaybooksAsTool(goctx gocontext.Context, srv *server.MCPServer, session server.ClientSession) error {
+	sessionID := session.SessionID()
```
yashmehrotra: The test suite server is stateless, so you might have to modify that if you want to test these

PR #2964, opened 2026-04-14
fix: infinite retries for notifications

adityathebe:
handle null retries

## Release Notes

* **Bug Fixes**
  * Enhanced notification retry logic to ensure consistent handling of max-retry thresholds for both pending and fallback notifications
  * Improved fallback notification triggering behavior to activate only when both fallback is explicitly configured and the retry limit is reached
  * Increased reliability of the notification delivery system through better edge case handling in retry progression


adityathebe:
fix here: https://github.com/flanksource/duty/pull/1885

PR #2998, opened 2026-04-21
fix: use default artifact connection for downloads when configured

adityathebe:
- Artifact downloads were always using `ctx.Blobs()`.
- This breaks old behavior when a default artifact connection is configured.
- Updated `DownloadArtifact` to use connection-backed filesystem reads (`GetFSForConnection`) when `DefaultArtifactConnection` is set.
- Falls back to `ctx.Blobs()` only when no default artifact connection is configured.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved artifact download reliability with enhanced error handling and connection management for artifact retrieval operations.

### DIFF
--- a/artifacts/controllers.go
+++ b/artifacts/controllers.go
@@ -17,6 +17,7 @@ import (
 	"github.com/labstack/echo/v4"
 
 	"github.com/flanksource/duty/rbac/policy"
+	icapi "github.com/flanksource/incident-commander/api"
 	"github.com/flanksource/incident-commander/db"
 	echoSrv "github.com/flanksource/incident-commander/echo"
 	"github.com/flanksource/incident-commander/rbac"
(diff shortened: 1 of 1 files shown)


moshloop:
@adityathebe shouldn't this be handled transparently by ctx.Blobs() ?

adityathebe:
> @adityathebe shouldn't this be handled transparently by ctx.Blobs() ?

ctx.Blobs() uses property. mission-control users a flag. we can standarize on one.

moshloop:
> > @adityathebe shouldn't this be handled transparently by ctx.Blobs() ?
> 
> ctx.Blobs() uses property. mission-control users a flag. we can standarize on one.

I think the flag should then just be a shortcut to setting the property and marked deprecated

adityathebe:
I'm realizing now that this PR wasn't necessary at all since we use the connection from the artifact record

PR #3051, opened 2026-05-04
Pr/plugins

moshloop:


api/v1/plugin_types.go
```diff
@@ -0,0 +1,135 @@
...
+	//+kubebuilder:validation:Required
+	Source string `json:"source"`
+
+	// Version of the binary to install. Forwarded to deps.Install verbatim
+	// (e.g. a semver tag, "latest", or a git ref the deps package supports).
+	Version string `json:"version,omitempty"`
+
+	// Checksum optionally pins the downloaded binary to a known good hash.
+	// When set, deps refuses to install a binary with a different checksum.
+	Checksum string `json:"checksum,omitempty"`
```
adityathebe, line 33: Probably need a plugin provider since we'll have to support remote plugins as well. And probably that'll be the primary use case anyway.

Here's a rough sketch.

```go

   type PluginProvider struct {
      // Downloads the binary and runs it
      Binary *BinaryPluginRuntime `json:"binary,omitempty"`

      // Talks to a remote plugin
      Remote *RemotePluginRuntime `json:"remote,omitempty"`
   }

   type BinaryPluginRuntime struct {
      Source   string `json:"source"`
      Version  string `json:"version,omitempty"`
      Checksum string `json:"checksum,omitempty"`

      // Optional override if needed later.
      Command []string `json:"command,omitempty"`
      Args    []string `json:"args,omitempty"`
      Env     map[string]string `json:"env,omitempty"`
   }

   type RemotePluginRuntime struct {
      // URL to the plugin gRPC/HTTP endpoint.
      Endpoint string `json:"endpoint"`

      // Optional connection/auth details.
      TLS      *PluginTLSConfig `json:"tls,omitempty"`
      Headers  map[string]string `json:"headers,omitempty"`
   }
```

api/v1/plugin_types.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// Selector decides which catalog (config) items this plugin's tabs
+	// attach to. The same ResourceSelector semantics used by Playbook.Configs
+	// apply: filter by config type, labels, tags, agent, namespace, name.
+	// An empty selector matches every config item.
+	//+kubebuilder:validation:Optional
+	Selector types.ResourceSelector `json:"selector,omitempty"`
```
adityathebe, line 40: I would put this under a `ui` field as we might almost certainly need to add more ui metadata later on.

api/v1/plugin_types.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// Connections is the set of connections the plugin may request at
+	// runtime. Reuses the same shape as Playbook actions[].exec.connections:
+	// typed fields per connection type (kubernetes/aws/gcp/azure) plus
+	// fromConfigItem, eksPodIdentity and serviceAccount flags. The plugin
+	// asks the host for a connection by type; the host resolves it from the
+	// matching field here (or the catalog item being viewed when
+	// fromConfigItem is set), and refuses any type the plugin didn't declare.
+	//+kubebuilder:validation:Optional
+	Connections connection.ExecConnections `json:"connections,omitempty"`
```
adityathebe: Should this be in the plugin spec or in permission CRD?

I think creating a permission is a better approach here.

adityathebe:
Merged in various other PRs

PR #3105, opened 2026-05-14
feat(plugins): add postgres connection status plugin

adityathebe:
Adds a Postgres plugin that resolves a configured Postgres connection and exposes a `connection_status` operation backed by `pg_stat_activity`.

The operation returns active, idle, and unknown connection counts while excluding the plugin's own backend session. It also includes a simple embedded UI tab that renders the counts as a pie chart with manual and periodic refresh.


adityathebe:
In the plugin registry

PR #3207, opened 2026-06-10
fix: update shutdown priority order 

yashmehrotra:
## Summary by CodeRabbit

* **Refactor**
  * Restructured server shutdown ordering so HTTP, plugin, worker, and database shutdown steps occur in a safer sequence. Added a cancellation step to ensure background workers stop before database connections are closed, improving graceful termination and resource cleanup and reducing risk of dangling work during shutdown.

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -18,7 +18,6 @@ import (
 	"github.com/flanksource/kopper"
 	"github.com/spf13/cobra"
 	"go.opentelemetry.io/otel"
-	ctrl "sigs.k8s.io/controller-runtime"
 
 	"github.com/flanksource/incident-commander/api"
 	v1 "github.com/flanksource/incident-commander/api/v1"
@@ -172,7 +171,7 @@ func launchKopper(ctx context.Context) {
 		ctx.Logger.Errorf("plugin replay from DB failed: %v", err)
 	}
 
-	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
+	if err := mgr.Start(ctx); err != nil {
 		shutdown.ShutdownAndExit(1, fmt.Sprintf("error running controller manager: %v", err))
 	}
 }
@@ -231,7 +230,7 @@ var Serve = &cobra.Command{
 		mcpServer := mcp.Server(ctx)
 		e.Any("/mcp", echov4.WrapHandler(mcpServer.HTTPHandler), mcp.AuthMiddleware)
 
-		shutdown.AddHookWithPriority("echo", shutdown.PriorityIngress, func() {
+		shutdown.AddHookWithPriority("echo", shutdown.PriorityCritical, func() {
 			echo.Shutdown(e)
 		})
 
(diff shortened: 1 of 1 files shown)


cmd/server.go
```diff
@@ -231,7 +231,7 @@ var Serve = &cobra.Command{
 		mcpServer := mcp.Server(ctx)
 		e.Any("/mcp", echov4.WrapHandler(mcpServer.HTTPHandler), mcp.AuthMiddleware)
 
-		shutdown.AddHookWithPriority("echo", shutdown.PriorityIngress, func() {
+		shutdown.AddHookWithPriority("echo", shutdown.PriorityCritical, func() {
```
adityathebe, line 234: Shouldn't echo be the first to shutdown?

cmd/server.go
```diff
@@ -244,17 +244,24 @@ var Serve = &cobra.Command{
...
 
-		shutdown.AddHookWithPriority("plugins", shutdown.PriorityIngress, func() {
+		shutdown.AddHookWithPriority("plugins", shutdown.PriorityJobs+1, func() {
 			machinery.StopAll(ctx)
 		})
 
-		shutdown.AddHookWithPriority("database", shutdown.PriorityCritical, stop)
+		// The database pool must close last (after workers have drained), but
+		// before echo unblocks main and the process exits.
+		shutdown.AddHookWithPriority("database", shutdown.PriorityCritical-1, stop)
```
adityathebe, line 253: With `PriorityCritical`, it should close after everything no?

cmd/server.go
yashmehrotra: No, it needs to be last because once the server stops the main process ends which then prevents the other shutdown goroutines to finish

cmd/server.go
yashmehrotra: yeah, highest priority gets killed last, this will get killed just before echo server

In general scenario (ie. k8s), once a pod is sent the termination command, it is also removed from the network (even if its running/in process of shutting down)

These changes are there to ensure graceful shutdown happens as much as possible

cmd/server.go
adityathebe: I see. In that case, Echo's shutdown should not exit the main process.

The first step during shutdown should be to stop accepting new requests. 
I don't think we want to have the http server running while the db is closed.

cmd/server.go
```diff
@@ -244,17 +244,24 @@ var Serve = &cobra.Command{
...
 
 		shutdown.WaitForSignal()
 
 		ctx.WithTracer(otel.GetTracerProvider().Tracer("mission-control"))
 		ctx = ctx.WithNamespace(api.Namespace)
 
+		// Cancel the context on shutdown so background workers (jobs, event
+		// consumers, kopper) stop before the database pool is closed.
+		ctx, cancel := ctx.WithCancel()
+		shutdown.AddHookWithPriority("workers", shutdown.PriorityJobs, func() { cancel() })
```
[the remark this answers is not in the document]
yashmehrotra: Invalid comment, WaitForSignal() spawns a goroutine

PR #3238, opened 2026-06-15
feat(faro): Allow multiple access tokens for the same server

### LINKED ISSUE #3235: faro: Cannot add multiple unique contexts for the same server
When adding a new context in presence of another existing context with the same server url, faro just reuses the same auth.

This prevents having different user context on the same server URL.

adityathebe:
resolves: https://github.com/flanksource/mission-control/issues/3235

Remove the need to store refreshtoken and id_token in a different file. It was adding an overhead to coordinate contexts with files.

## Release Notes

* **Refactor**
  * Streamlined authentication token management by consolidating storage within the context configuration system.
  * OIDC tokens are now persisted alongside context data instead of separate files, improving consistency.
  * Enhanced token resolution to automatically refresh credentials when approaching expiration.
  * Unified token access across CLI commands through improved context integration.

### DIFF
--- a/clientcmd/api_client.go
+++ b/clientcmd/api_client.go
@@ -3,10 +3,10 @@ package clientcmd
 import (
 	"context"
 	"fmt"
-	"os"
 	"sync"
 
 	"github.com/flanksource/commons/logger"
+	"github.com/flanksource/incident-commander/auth/oidcclient"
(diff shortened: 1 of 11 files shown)


adityathebe:
@coderabbitai review

PR #3300, opened 2026-06-29
feat: add MCP whoami tool

adityathebe:
Adds a read-only MCP `whoami` tool so clients can inspect the current authenticated subject.

The tool returns the authenticated user/token, delegated owner, delegation status, roles, permissions, and MCP access status. This helps MCP clients/debugging flows verify identity without relying on prompt injection.


adityathebe:
Needed it for a test, but I can use faro whoami

PR #3334, opened 2026-07-07
fix(playbook): align scheduled runs with shared pipeline

### LINKED ISSUE #3305: Scheduled Playbook Missing Validations
| issue                            | description                                                                                   |
| -------------------------------- | --------------------------------------------------------------------------------------------- |
| parameters never templated       | Declared template:"true" + documented as such, but stored raw → implement or stop claiming it |
| Cron string never validated      | Invalid cron saves clean, fails silently every minute                                         |
| No parameter validation/defaults | Required params unenforced; runs created broken                                               |

### LINKED ISSUE #3306: Scheduled Playbook ignores `approval` (silent discrepancy across trigger paths)
### Summary

A playbook with `spec.approval` defined behaves differently depending on how it is triggered. The **schedule** trigger silently ignores the approval block.

| Trigger path | Code | Approval honored? |
| --- | --- | --- |
| Manual / API run | `playbook/playbook.go:126-127` | ✅ sets `PendingApproval` |
| Event (config/canary/component) | `playbook/events.go:124-125` | ✅ sets `PendingApproval` |
| **Schedule** | `playbook/schedule.go:134-140` | ❌ hard-codes `Status: PlaybookRunStatusScheduled` |

`triggerScheduledRun` creates the run directly in `Scheduled` status. The run consumer (`playbook/run_consumer.go:369-373`) executes anything in `Scheduled`, and `runner.ScheduleRun` never re-checks approval. So a scheduled playbook that also has `approval` runs **without** approval.

### Why this matters

The danger is the *silent asymmetry*: a playbook authored with approval (enforced on the manual/event paths) that later gets a `schedule:` block will quietly run unapproved on schedule, with no signal to the author.

### Resolution

Per-run approval on a recurring schedule is not a sensible design — a `@every 5m` playbook would manufacture a `pending_approval` run every 5 minutes that no approver can realistically clear; they'd just pile up (the consumer only picks `Scheduled`). No external scheduler (k8s CronJob, Argo, EventBridge…) gates per-tick either.

So the fix is **not** to make scheduled runs go `PendingApproval`. The combination is contradictory and should be **rejected at validation time** in `PlaybookSpec.Validate()` (`api/v1/playbook_types.go:355`) — fail at apply with a clear message rather than silently dropping the approval block.

Related: #3305

### LINKED ISSUE #3307: Scheduled Playbook ignores `spec.timeout`
### Summary

Scheduled runs always use the global default timeout and ignore the playbook's own `spec.timeout`.

### LINKED ISSUE #3308: Scheduled playbook trigger failures are not surfaced (log-only)
When a scheduled playbook fails to fire or create a run, the only trace is a log line. Nothing is recorded to job_history, retried, or made queryable.

By contrast, the event path records such failures to job_history via logToJobHistory ([playbook/events.go:166-174](https://github.com/flanksource/mission-control/blob/main/playbook/events.go#L166-L174)), keyed to the playbook, which surfaces in the existing job-history UI.

adityathebe:
Scheduled playbook runs were created directly from the scheduler path, bypassing the normal run preparation used by manual/webhook runs.

This meant scheduled runs skipped parameter defaulting/validation, timeout parsing, spec validation, and failure visibility in job history.

Route scheduled runs through a shared internal run creation helper, template schedule parameters before validation, and record scheduler run creation failures in job history. Also reject invalid schedule specs and schedule+approval configs up front.

Resolves #3305
Resolves #3306
Resolves #3307
Resolves #3308

## Summary by CodeRabbit

* **New Features**
  * Added validation for cron-based schedules, including detection of invalid expressions.
  * Scheduled playbooks now support templated parameters and updated parameter values at execution time.
  * Scheduled executions verify that the configured schedule remains current before running.

* **Bug Fixes**
  * Prevented scheduled playbooks requiring approval from being accepted.
  * Improved failure reporting by recording unsuccessful scheduled jobs in job history.
  * Applied configured defaults and timeouts consistently to scheduled runs.

### DIFF
--- a/api/v1/playbook_types.go
+++ b/api/v1/playbook_types.go
@@ -13,6 +13,7 @@ import (
 	"github.com/flanksource/kopper"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/uuid"
+	"github.com/robfig/cron/v3"
 	"github.com/samber/lo"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -380,6 +381,19 @@ func (p PlaybookSpec) Validate() error {
 			return err
 		}
 	}
+
+	if p.On != nil && len(p.On.Schedule) > 0 {
+		if p.Approval != nil && !p.Approval.Approvers.Empty() {
+			return fmt.Errorf("scheduled playbooks cannot require approval")
+		}
+
+		for i, schedule := range p.On.Schedule {
+			if _, err := cron.ParseStandard(schedule.Schedule); err != nil {
+				return fmt.Errorf("invalid schedule[%d] %q: %w", i, schedule.Schedule, err)
+			}
+		}
+	}
+
 	return nil
 }
 
--- a/playbook/playbook.go
+++ b/playbook/playbook.go
@@ -94,14 +94,12 @@ func CreateOrSaveFromFile(ctx context.Context, file string) (*models.Playbook, e
 
 type playbookRunOptions struct {
 	CheckPermissions bool
-	AllowApproval    bool
 }
 
 // Run creates and saves a run from a run request after validating the run parameters.
 func Run(ctx context.Context, playbook *models.Playbook, req RunParams) (*models.PlaybookRun, error) {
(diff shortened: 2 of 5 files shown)


adityathebe:
@codex[agent] Please Review

PR #3456, opened 2026-09-02
fix(mail): load system SMTP from DB on every send

### LINKED ISSUE #3455: SMTP cache remains stale after connection updates
After the system email connection is updated or deleted, Mission Control can continue using the previous SMTP configuration for up to 15 minutes. Outgoing email may use stale credentials or server settings instead of the saved connection.

References

- https://github.com/flanksource/mission-control/commit/a106d06e9158bd8feefb66166221e93e1d960ab2

adityathebe:
Closes #3455

Updating or deleting the system email connection left Mission Control using the previous SMTP host/credentials for up to 15 minutes.

`GetDefaultSMTP` cached that row in process memory. Flushing depended on a `connections` `table_activity` notify, but Duty never installs that trigger, so the cache never cleared.

Drop the cache and load the current system email connection on each send. Email notifications and invite mail now pick up connection changes immediately.

## Summary by CodeRabbit

* **Bug Fixes**
  * SMTP configuration is now loaded from the database each time it is requested, ensuring current settings are used.
  * Environment variables and command-line settings remain available as fallbacks when no database configuration is found.
  * SMTP settings are no longer served from a potentially stale in-memory cache.

### DIFF
--- a/cmd/server.go
+++ b/cmd/server.go
@@ -29,7 +29,6 @@ import (
 	"github.com/flanksource/incident-commander/echo"
 	"github.com/flanksource/incident-commander/events"
 	"github.com/flanksource/incident-commander/jobs"
-	"github.com/flanksource/incident-commander/mail"
 	"github.com/flanksource/incident-commander/mcp"
 	"github.com/flanksource/incident-commander/metrics"
 	"github.com/flanksource/incident-commander/notification"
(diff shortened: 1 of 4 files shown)


moshloop:
@adityathebe - I think we should rather fix the connections table_activity trigger, as its more cross-cutting and doesn't involve a secret read in the critical path (i.e. k8s being under load and unable to return an existing secret shouldn't block the notification of k8s being under load)

PR #3457, opened 2026-09-03
fix: prevent pg_notify cache fills from restoring stale values

### LINKED ISSUE #3454: pg_notify cache invalidation can be undone by in-flight cache fills
Several in-process caches (notifications by event/ID, team specs, view scopes, RLS payloads and PostgREST JWTs, SMTP config, and duty's entity getter cache) are invalidated from `table_activity` pg_notify events. If a cache miss is being filled from the database while the invalidation arrives, the stale result is stored after the purge and stays cached until its TTL expires (up to 12 hours for notifications, 1 hour for RLS/JWT tokens). This means changes to notifications, teams, scopes, or permissions can intermittently fail to take effect for a long time even though the invalidation notification was delivered.

## References

- #3452 (same race observed in the playbook event cache)
- #3453

adityathebe:
`table_activity` invalidation could be undone when a cache miss was already in flight: the handler purged an empty key, then the stale SELECT result was stored until TTL (up to 12 hours for notifications, 1 hour for RLS/JWT). Permission and notification changes could therefore fail to take effect even though the notify was delivered.

Folds a generation counter into cache keys via a shared helper so `Flush` and `Delete` make in-flight `SetIf` writes invisible. Per-user RLS invalidation now flushes the shared token cache (impersonated payloads and PostgREST JWTs share it), team-member invalidation runs after RBAC reload, and team specs are purged on insert as well as update/delete.

Fixes #3454

Duty's `getterCache` still has this race inside `query.findEntityByField`; that needs a Duty-side change and a dependency bump.

### DIFF
--- a/auth/rls.go
+++ b/auth/rls.go
@@ -18,6 +18,7 @@ import (
 	"gorm.io/gorm"
 
 	v1 "github.com/flanksource/incident-commander/api/v1"
+	"github.com/flanksource/incident-commander/utils"
 	"github.com/flanksource/incident-commander/vars"
 )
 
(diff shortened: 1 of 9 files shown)


adityathebe:
Superseded by smaller stacked PRs: playbook event cache first, then remaining fill sites.

